PR #8, opened 2015-07-29
Removed THREE.Math.degToRad dependency from gyro sensor.

craigsketchley:


borismus:
The main THREE dependency is for Euler, Quaternion, Vector3, etc. What's the motivation behind this change?

craigsketchley:
Sorry, thought I wrote a description. I wasn't using THREE.js in my project and the polyfill currently references `THREE.Math.degToRad` in `gyro-position-sensor-vr-device.js`. I get an exception when I try to run my project on mobile.

borismus:
Ah, a very good point. Maybe a better approach is to patch in THREE.Math.degToRad in the same way that I do with the other THREE.* classes?

borismus:
At any rate would appreciate if you could rebase the patch to the latest master. Thanks!

borismus:
This has been fixed. We now pull in THREE.Math.*

PR #11, opened 2015-09-02
Cardboard HMD: Provide skewed FOVs to simulate a "focal distance"

njam:
Via https://github.com/mrdoob/three.js/pull/7036#issuecomment-135509274

I understand the FOVs for each eye should be slightly skewed inwards to simulate a focal distance.
See for example here: http://paulbourke.net/stereographics/stereorender/

The values I'm using are the ones provided by the experimental Chrome WebVR builds.
They result in a focal distance of ~1.90.

I've also changed the interpupillary distance to the average [according to Wikipedia](https://en.wikipedia.org/wiki/Interpupillary_distance).

wdyt?
cc @tschw

### DIFF
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -60,20 +60,27 @@ var HMDVRDevice = require('./base.js').HMDVRDevice;
 
 // Constants from vrtoolkit: https://github.com/googlesamples/cardboard-java.
 var INTERPUPILLARY_DISTANCE = 0.06;
-var DEFAULT_MAX_FOV_LEFT_RIGHT = 40;
-var DEFAULT_MAX_FOV_BOTTOM = 40;
-var DEFAULT_MAX_FOV_TOP = 40;
+var DEFAULT_MAX_FOV_TOP = 53.09438705444336;
+var DEFAULT_MAX_FOV_BOTTOM = 53.09438705444336;
+var DEFAULT_MAX_FOV_OUTER = 46.63209533691406;
+var DEFAULT_MAX_FOV_INNER = 47.52769470214844;
 
 /**
  * The HMD itself, providing rendering parameters.
  */
 function CardboardHMDVRDevice() {
   // From com/google/vrtoolkit/cardboard/FieldOfView.java.
-  this.fov = {
+  this.fovLeft = {
     upDegrees: DEFAULT_MAX_FOV_TOP,
     downDegrees: DEFAULT_MAX_FOV_BOTTOM,
-    leftDegrees: DEFAULT_MAX_FOV_LEFT_RIGHT,
-    rightDegrees: DEFAULT_MAX_FOV_LEFT_RIGHT
+    leftDegrees: DEFAULT_MAX_FOV_OUTER,
+    rightDegrees: DEFAULT_MAX_FOV_INNER
+  };
+  this.fovRight = {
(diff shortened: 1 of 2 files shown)


borismus:
I just tried this in Cardboard on a Nexus 5, and it looks super weird to me.

tschw:
Point is, the device should use a viewport that is focused, ideally on a depth that looks non-weird :-). The calculation to skew the "all 40s angles" for the two images to converge at a certain distance is:

```javascript
var TOTAL_FOV = 80;
var FOCAL_DISTANCE = 1.5; // perceived distance in meters

var TAN_HALF_FOV = Math.tan(TOTAL_FOV * Math.PI / 360); 
var SKEW_SLOPE = INTERPUPILLARY_DISTANCE / (FOCAL_DISTANCE * 2);

var DEFAULT_MAX_FOV_TOP = TOTAL_FOV / 2;
var DEFAULT_MAX_FOV_BOTTOM = TOTAL_FOV / 2;
var DEFAULT_MAX_FOV_OUTER = Math.atan(TAN_HALF_FOV - SKEW_SLOPE);
var DEFAULT_MAX_FOV_INNER = Math.atan(TAN_HALF_FOV + SKEW_SLOPE);
```

Just pick a reasonable default for `FOCAL_DISTANCE` based on what you see.

borismus:
How do you avoid toe-in described as undesirable in http://paulbourke.net/stereographics/stereorender ?

tschw:
There's nothing we can do about it here - it depends on how the application (or underlying 3D library) constructs the matrices: 

A rotation by the average of the left/right angles gives toe-in, instead we want to shear x along the z axis by what I called `SKEW_SLOPE` above.

This means the `left` and `right` parameters of [`glFrustum`](https://www.opengl.org/sdk/docs/man2/xhtml/glFrustum.xml)-style projection matrices must not have the same absolute value, but be asymmetric in respect to the origin; they express the bounds of the near plane. Adding the same distance to both `left` and `right` yields a skewed frustum. That is, it extends as if we were looking sideways through a window (note that there is no rotation, z remains orthogonal to the projection plane).
At a certain distance, the translation of the cameras (not part of a bare projection matrix) must be nullified by the skew, and rendering an x/y-plane at that distance yields exactly the same image for both eyes. This is the focal distance. For 3D monitors, it's what's perceived *on* the screen - as opposed to before or behind it. 

Now, the first element of the `glFrustum` matrix, the x scale, is actually a quotient of the near distance and values specified at that distance. The scale doesn't care at which distance we determine its value, as long as the scheme remains consistent. Setting `zNear = 1` actually makes `left` and `right` the same as the tangents of the WebVR angles. This approach works for many of the values in the matrix, so we can pluck the tangents directly into it. That's what the Three.js code does:

https://github.com/mrdoob/three.js/blob/master/examples/js/effects/VREffect.js#L173-L181
https://github.com/mrdoob/three.js/blob/master/examples/js/effects/VREffect.js#L196-L222

Note that your scene should use meters as world unit or `.scale` must be set accordingly. 

We still want to improve it, such as steer the focus, know the exact position of the focal plane for sharp 2D rendering, allow input from authoring formats, and have one controller for stereoscopic rendering that works for all possible output methods, respecting the constraints from the min/max values in case it's WebVR.

borismus:
This should now be resolved in https://github.com/borismus/webvr-boilerplate/commit/363840e07e05375159c998f580db162c3c0cd4d2.

PR #12, opened 2015-09-29
Free wasd key for camere movement

wizztjh:
We are using wasd to move the camera around. We need to disable the wasd keycode binding. What do you guys think?


borismus:
Fine with me. WASD are pretty commonly needed for non-camera stuff. Thanks!

PR #13, opened 2015-10-22
Device orientation fixes for Firefox Android

caseyyee:
Adds support for landscape device orientation with Firefox Android.

* uses window resize and mediaquery list object to detect orientation in absence of having orientationchange events.
* swaps orientationevent axis' when in portrait orientation.


borismus:
Sorry to pull the rug from under you, but moving to a sensor fusion model, which is much more performant. New implementation relies on `devicemotion` events. I tested a bit with FF for Android, and the sensor data seems to be coming in really slowly. Any ideas why?

caseyyee:
@borismus Just filed a bug for that here: https://bugzilla.mozilla.org/show_bug.cgi?id=1217942
Nothing we can really do about that until there is a fix in platform for the issue.

Just so that it's serviceable in Firefox Android, the Polyfill should fallback to deviceOrientation with FF.   Will update this PR.

borismus:
Thanks!

caseyyee:
There are patches inbound for `devicemotion` and `orientationchange` into Firefox.  Cancelling PR and will work on patches for the fusion sensor device instead.

PR #14, opened 2015-10-30
Fusion sensor device compatibility with Firefox Android.

caseyyee:
Firefox Android Nightly edition has now landed faster devicemotion event updates,  windoworientationchange events and window.orientation property, which means that we can take advantage of the fusion position sensor device.

This changes in this PR account for the remaining differences in how Firefox returns devicemotion event timestamp.


src/util.js
```diff
@@ -14,7 +14,7 @@
  */
 var Util = window.Util || {};
 
-Util.MIN_TIMESTEP = 0.001;
+Util.MIN_TIMESTEP = 0.0001;
```
borismus: Why?

src/fusion-position-sensor-vr-device.js
```diff
@@ -111,7 +119,7 @@ FusionPositionSensorVRDevice.prototype.onDeviceMotionChange_ = function(deviceMo
 
   // In iOS, rotationRate is reported in degrees, so we first convert to
```
borismus: Update comment too, please.

src/util.js
caseyyee: Yes, we seem to be getting a high resolution timestamp.

borismus: Very odd. Seems from your CL like FF gives timestamps in microseconds, but you convert to milliseconds. It seems impossible that your update rate is faster than the new timestep (which is 1/0.0001=10,000 Hz), since phone firmware only gives sensors at 200 Hz at most.

Could you investigate?

caseyyee: Boris,
I did take a look at this and it seems that FF returns smaller deltas.   Not sure why this is the case.   

I whipped together a page to monitor the differences: https://github.com/caseyyee/DevicemotionTest

caseyyee: It looks like there are different rates that can be set for reading out Android sensor data at: NORMAL, UI, GAME or FASTEST.   Firefox devicemotion events are set to use _Fastest_.

For my device, the _Fastest_ rate is about 120hz, which equals to 9ms.   So what I am getting back seems about right for Firefox.  

Maybe Chrome uses _Game_ rate?   Which on my device is ~50hz, which is 20ms.   Seems a bit off to me, but maybe still within the range?

PR #16, opened 2015-11-06
add npm `package.json`

cvan:


borismus:
Why? I'd like to stick to one package system: bower.

cvan:
> Why? I'd like to stick to one package system: bower.

Understood - package managers, build scripts, they can complicate things. Admitteldly, being able to `npm install webvr-polyfill` is nice. The [`webvr-polyfill` package](https://www.npmjs.com/package/webvr-polyfill) on npm is at version 0.1.1. Would it be possible to update it? Or if it's not going to be published on npm, would it be possible to add a message in the README for the npm version or unpublish?

borismus:
Could you bower install it?

dmarcos:
```npm``` and ```bower``` are extremely popular. I would even say that npm is taking the lead at the moment for front end modules. Publishing in both carries a very low dev cost and will help the adoption of your library. You don't enforce a specific tool to your users.

borismus:
If all it takes is a version bump in the .json file, I guess that's not so onerous.

bnolan:
Thanks, I use npm for building scenevr and having an up to date npm package is very helpful, cheers!

PR #18, opened 2015-11-09
Flag to disable touch panner in fusion sensor mode

dmarcos:
I'm using the head tracking functionality and it works great. On the same project I'm also implementing my own touch controls and the builtin `touch panner` conflicts with them. I would like to have a flag to disable them.


borismus:
Thanks!

PR #20, opened 2015-11-15
Flag to disable MouseKeyboardPositionSensorVRDevice

dmarcos:
This is similar to my previous https://github.com/borismus/webvr-polyfill/pull/18
I have my own mouse and keyboard controls and ```MouseKeyboardPositionSensorVRDevice``` is conflicting with them. I would like to have the option to disable them. If you want to handle this differently let me know and I will update the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,4 +50,6 @@ are supported:
       //PREDICTION_TIME_S: 0.050, // Default: 0.050s.
       // Flag to disable touch panner. In case you have your own touch controls
       //TOUCH_PANNER_DISABLED: true, // Default: false.
+      // To disable keyboard and mouse controls. In case you implement your own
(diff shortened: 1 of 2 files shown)


borismus:
I'm curious what you're doing differently in your project? Are you just using OrbitControls? What customization are you seeking that MouseKeyboardPositionSensorVRDevice doesn't work for you.

dmarcos:
Your controls prescribe a specific type of locomotion that is not suitable for all experiences. For instance, controlling a walking avatar is not the same as controlling a car or a flying vehicle. It's very nice that you provide a default that get you going but you should allow people disable them and implement their control schemes.

borismus:
Fair enough!

PR #31, opened 2016-01-26
Add Y-axis Control for TouchPanner.

jun784:
I added Y-axis touch control with TouchPanner.

http://recordit.co/CTV4bxB24Z


borismus:
I don't think we actually want this. Panning arbitrarily in Y, you can end up in really weird states. If you want this, I'd suggest a maximum absolute pitch.

PR #37, opened 2016-03-04
Make a wrapper object fullscreen instead of body

toji:
Making the body fullscreen doesn’t work if the canvas is not the only
element on the page. Elements like buttons, text, or images may end up
overlaid on the distorted output.

This patch injects a wrapper div into the page, moves the canvas into
it, and makes that fullscreen instead. That way the content  of the
wrapper div can be tightly controlled and no unintended elements will
be displayed.


src/base.js
```diff
@@ -61,11 +64,49 @@ VRDisplay.prototype.cancelAnimationFrame = function(id) {
...
+VRDisplay.prototype.wrapForFullscreen = function(element) {
+  if (Util.isIOS())
+    return element;
+
+  if (!this.fullscreenWrapper_) {
+    this.fullscreenWrapper_ = document.createElement('div');
+    this.fullscreenWrapper_.classList.add('webvr-polyfill-fullscreen-wrapper');
+  }
+
+  if (this.fullscreenElement_ == element)
```
borismus: Nit: braces all ifs

src/base.js
```diff
@@ -61,11 +64,49 @@ VRDisplay.prototype.cancelAnimationFrame = function(id) {
...
+  var parent = this.fullscreenElement_.parentElement;
+  parent.insertBefore(this.fullscreenWrapper_, this.fullscreenElement_);
+  parent.removeChild(this.fullscreenElement_);
+  this.fullscreenWrapper_.insertBefore(this.fullscreenElement_, this.fullscreenWrapper_.firstChild);
+
+  return this.fullscreenWrapper_;
+};
+
+VRDisplay.prototype.removeFullscreenWrapper = function() {
+  if (!this.fullscreenElement_)
```
borismus: ditto

borismus:
Thanks!

PR #43, opened 2016-03-14
Added back button to cardboard UI

toji:
It's already wired up to call VRDevice.exitPresent when clicked. Also rebuilt the build/ version which apparently picked up some older changes.


borismus:
Thanks! Any luck with the iOS scaling issue?

toji:
Need to dig out my old iOS devices and test it still.

Issue #46, opened 2016-03-22
Fix UI in iOS


borismus:
Some UI glitches in iOS.

- White background (around the eyes) should be black in VR mode
- Back button should be black, and much smaller
- Back button should work

![img_1180](https://cloud.githubusercontent.com/assets/277632/13967199/32cf22bc-f033-11e5-80e3-5fdca1e7b69c.PNG)

cc: @toji

cvan:
I noticed these other issues:

- The back button is too large on Android (not as large as on iOS). At least on my devices, the button obscures the top corner of the left viewport.
- The gear icon is noticeably larger on iOS than on Android.
- Clicking on the gear icon doesn't do anything on Android (or anywhere where the Fullscreen API is supported). When I exit VR, for a few milliseconds before it disappears, I can see that the Cardboard settings modal was behind the fullscreened canvas. On iOS the menu appears because WebKit doesn't support Fullscreen.

Not sure whether you want to cover all of those issues in this one or file separate ones. Let me know - I can do that.

borismus:
Thanks for testing more. Are you using webvr-samples, or using the latest
polyfill? I thought I fixed the gear issue (your #3) but I don't think
@toji updated to latest polyfill yet.

cvan:
Latest polyfill

borismus:
Things are looking a little bit better here: http://borismus.github.io/webvr-polyfill/examples/basic/. There's still a problem on iOS, which manifests when you exit VR mode. I'm not sure how to best fix it... Problem is the following:

When we go into VR mode, we redefine canvas.width and canvas.height using `Object.defineProperty` (https://github.com/borismus/webvr-polyfill/blob/master/src/cardboard-distorter.js#L238). When exiting VR mode, we restore canvas.width and canvas.height with its original values, which we capture using `Object.getOwnPropertyDescriptor` (https://github.com/borismus/webvr-polyfill/blob/master/src/cardboard-distorter.js#L68). However on iOS, these values never really get set back properly, so changing canvas.width/height values has no effect.

cvan:
I noticed the exact same thing; spent a few hours last night trying to fix but I was unsuccessful. maybe y'all have better luck.

cvan:
Testing against `examples/basic/`, I noticed that the back button is smaller, but it's still visible within the left viewport. And the back button on iOS seems to make the screen turn black and rotating the phone back to portrait doesn't seem to resize things correctly.

borismus:
[...]

PR #56, opened 2016-03-27
Taking a stab at fixing some iOS issues.

toji:
I don't have an iOS device with me to test this with, sadly, but I think it should avoid the problems we're seeing in Issue #46. (I was at least able to test it on desktop Safari.) It no longer patches the canvas width and height in Safari and instead checks the values each `submitFrame` and resizes the framebuffer when needed. This will cause some minor quality and perf issues but prevent larger correctness ones. Also forces the backbuffer to clear every `submitFrame`, which should prevent the white background we're seeing. I also took the liberty of doing a minor optimization to the `Util.is...` functions, since they're going to get called a lot more frequently now and their return value should never change.

*Please test this on an iOS device before merging!*

### DIFF
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -982,6 +982,7 @@ var nextDisplayId = 1000;
  * The base class for all VR displays.
  */
 function VRDisplay() {
+  this.isPolyfilled = true;
   this.displayId = nextDisplayId++;
   this.displayName = 'webvr-polyfill displayName';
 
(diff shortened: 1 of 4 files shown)


borismus:
Thanks, I tested this a bit (merged it on to my gh-pages branch which includes a basic example). Have you tried it with the iOS simulator? I'm seeing a sizing issue when exitPresent happens.

screen shot 2016-03-27 at 12 32 57 pm

borismus:
Also... the button size is still screwy. 

screen shot 2016-03-27 at 2 44 49 pm

borismus:
Ok, problem gets resolved when I reset the viewport on the THREE.js renderer:

`renderer.setViewport(0,0,width,height)`

I've worked around the issue by having the sample resize when we change VR presentation modes:

`window.addEventListener('vrdisplaypresentchange', onResize);`

PR #58, opened 2016-03-29
add node modules + add `three` to package for basic example

cvan:
per the PR I opened last week (#34), I still recommend not checking in `node_modules`. it can complicate things such as rebasing, merging, and bisecting - especially if developers of the project are using different operating systems or versions of npm (i.e., v2 vs. v3 which flattens dependencies).

but, if you want to keep `node_modules` checked in, we should probably commit the `node_modules` directory (for `es6-promise`) to the root of this project.

there's already a `node_modules` for `examples/basic/`, so didn't need to change anything there.


borismus:
I don't think we need all of the devdependencies present locally in the sample. Just three.js. I've updated the package.json with the three.js dependency, also added the es6 dep.

cvan:
what is the workflow you recommend?

1. clone a repo.
2. `npm install`
3. `npm start` (or in this case: `npm run watch`)
4. open files (or start a simple server such as `http-server -c-1`)

not for the examples, but how else would folks dev against the `webvr-polyfill` code in the `src/` directory? that's where I'm unclear.

borismus:
My workflow is this. Clone, (npm should be updated), npm run watch. Edit src/*, things get rebuilt automatically. Workflow: reload `http://localhost/~smus/Projects/webvr-polyfill/examples/basic/` (I have apache running locally permanently).

cvan:
@borismus okay but what about when you're making changes to the polyfill in `src/`? because currently `node_modules` is not checked into the repo, so you'd have to run `npm install`. when you run `git status -u`, do you not see all the unchecked `node_modules`?

borismus:
Not sure what you mean. All we need is three.js, which is checked in here:
https://github.com/borismus/webvr-polyfill/tree/master/examples/basic

cvan:
> Not sure what you mean. All we need is three.js, which is checked in here:

[`es6-promise` is included in the `package.json`](https://github.com/borismus/webvr-polyfill/blob/master/package.json#L11) and is used in [`src/main.js`](https://github.com/borismus/webvr-polyfill/blob/8fef7b/src/webvr-polyfill.js#L16-L17)

cvan:
`es6-promise` was added in 2b225de (Mar 2) and as of 4a7ef5b (Mar 29), it wasn't included. it looks like 7e81bfe (Mar 29) you added `es6-promise` to `node_modules`.

so we're good now :smile:

I discovered my issue though (npm v3) and I'll file separately.

PR #64, opened 2016-04-05
do not vendor `node_modules` (fixes #62)

### LINKED ISSUE #62: Do not vendor `node_modules` for `src/`
I mentioned this in a couple other issues/PRs (such as #34), [this one most recently](https://github.com/borismus/webvr-polyfill/pull/58#issue-144127095) for #58.

I recommend reconsidering "vendoring" the [`node_modules`](https://github.com/borismus/webvr-polyfill/tree/master/node_modules) directory for the [main lib code (`src/`)](https://github.com/borismus/webvr-polyfill/tree/master/src).

I think committing the `node_modules` for the `examples/` is fine mainly for these two reasons:

1. **Works (and deployable) out of the box.** folks can fire up the [`examples/`](https://borismus.github.io/webvr-polyfill/examples/basic/) locally (and when deployed to [GitHub Pages](https://borismus.github.io/webvr-polyfill/examples/basic/)) without needing to first run any `npm` scripts (assuming `build/webvr-polyfill.js` is up to date).
2. **Contains external dependencies that are not included in the `webvr-polyfill` Browserify bundle.** [`examples/basic/index.html` relies on JS files that are not included in the `build/webvr-polyfill.js` bundle](https://github.com/borismus/webvr-polyfill/blob/ff3d2d6/examples/basic/index.html#L54-L61). 

For the main library code in [`src/`](https://github.com/borismus/webvr-polyfill/tree/master/src), however, I think [its `node_modules`](https://github.com/borismus/webvr-polyfill/tree/master/node_modules) should not be checked in. Right now, the only dependency for `webvr-polyfill` bundle is `es6-promise`, but I can see that changing soon - perhaps for a few of these modules:

* [`THREE.Math`](https://github.com/borismus/webvr-polyfill/blob/master/src/three-math.js) (discussed in issue #2).
* [`headtrackr`](https://github.com/borismus/webvr-polyfill/blob/master/src/deps/headtrackr.js)
* [`wglu-preserve-state`](https://github.com/borismus/webvr-polyfill/blob/master/src/deps/wglu-preserve-state.js)
* [`cardboard-dpdb`](https://github.com/borismus/webvr-polyfill/tree/master/src/dpdb)

Anyway, off the top of my head, these are the main reasons I could think of to stop vendoring `src/`'s `node_modules`:

0. **Forces developers to use npm2 or npm3.** npm3 introduced a flat dependency-tree structure. This means whichever version of npm was used to commit the `node_modules` must match the user's npm version. Many folks, myself included, have switched over to npm3 (its official release was back in June 2015); and, for what it's worth, it's also the version installed by Homebrew now when a user runs `brew install npm`. This causes a mess of unstaged dependencies when trying to commit code:
screenshot 2016-04-03 22 57 33
    You could upgrade to npm3 and commit the `node_modules` with the new npm3 directory structure. But that also means any users using npm2 will be out of luck and will be forced to upgrade to npm3.
0. **`npm install` is easy for contributors.** since npm scripts are necessary for doing local development, I have to already [`npm run build`](https://github.com/borismus/webvr-polyfill/blob/master/package.json#L25) and [`npm run watch`](https://github.com/borismus/webvr-polyfill/blob/master/package.json#L26), I think it's safe to say that developers will have no issue running `npm install` once after cloning the repo (and with new updates, of course).
0. **Reviewing pull requests with vendor changes are messy, noisy, and potentially unsafe.** When third-party modules are added or upgraded, malicious code changes may go undetected in the large diffs. Especially on GitHub, the diffs in the PRs won't even be visible; you'd have to apply the patch locally to see the `git diff`. It's definitely an aside, but calculating hashes of the `node_modules` is really the only way to ensure what's obtained from `npm` is what's in `node_modules`. And [`npm shrinkwrap`](https://docs.npmjs.com/cli/shrinkwrap) is the solution for that. I've used it in the past, and it's helpful especially when you want to make sure that the dependencies used in your local development environment are the exact same ones used in your production environment (mostly important for backend code of course).
0. **Requires careful, manual work to remove dependencies.** Folks should always be removing dependencies with `npm uninstall`, but many folks don't; they instead just remove it from the `package.json`. In these cases, if not caught in the pull request, an unused dependency could easily linger in the vendored `node_modules` directory. And with npm3's flat dependency-tree structure, the dependency relationships are no longer obvious, so it's usually nearly impossible to discern which modules are still needed by looking at the `node_modules`. This is a solved problem if the dependencies are instead managed by the user with `npm install`.
0. **Lack of precedent in OSS libraries.** Although I have heard of folks "vendoring" their `node_modules`, I actually can't find any examples of OSS Node projects that do it.

Apologies for the verbosity. It's not my intent to be pedantic; I wanted to be thorough in explaining the case against vendoring `src/`'s `node_modules`. This was actually a good exercise to think of all the pros and cons. Anyway, let me know your thoughts.

cvan:


borismus:
Ok, you convinced me.

PR #73, opened 2016-04-06
Use whatcardboard iframe for viewer selector

stuartpb:
This allows users to pick from all known Google Cardboard headsets (not just the Google I/O ones), or scan their headset's QR code (memory permitting), and have that choice be remembered across all
sites using the WebVR polyfill (not just the current domain).

### DIFF
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -1,4 +1,97 @@
 (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(_dereq_,module,exports){
+// shim for using process in browser
+
+var process = module.exports = {};
+var queue = [];
+var draining = false;
+var currentQueue;
+var queueIndex = -1;
+
+function cleanUpNextTick() {
+    draining = false;
+    if (currentQueue.length) {
+        queue = currentQueue.concat(queue);
+    } else {
+        queueIndex = -1;
+    }
+    if (queue.length) {
+        drainQueue();
+    }
+}
+
+function drainQueue() {
+    if (draining) {
+        return;
+    }
+    var timeout = setTimeout(cleanUpNextTick);
+    draining = true;
+
+    var len = queue.length;
+    while(len) {
+        currentQueue = queue;
+        queue = [];
+        while (++queueIndex < len) {
+            if (currentQueue) {
+                currentQueue[queueIndex].run();
+            }
+        }
+        queueIndex = -1;
+        len = queue.length;
+    }
+    currentQueue = null;
+    draining = false;
+    clearTimeout(timeout);
+}
+
+process.nextTick = function (fun) {
+    var args = new Array(arguments.length - 1);
+    if (arguments.length > 1) {
+        for (var i = 1; i < arguments.length; i++) {
+            args[i - 1] = arguments[i];
+        }
+    }
+    queue.push(new Item(fun, args));
+    if (queue.length === 1 && !draining) {
+        setTimeout(drainQueue, 0);
+    }
+};
+
+// v8 likes predictible objects
+function Item(fun, array) {
+    this.fun = fun;
+    this.array = array;
+}
+Item.prototype.run = function () {
+    this.fun.apply(null, this.array);
+};
+process.title = 'browser';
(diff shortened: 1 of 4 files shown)


stuartpb:
This removes the `inverseCoefficients` fields on headset data, which was, as far as I (and @toji) could tell, unused; #74 introduces a function that would ameliorate this, if needed.

stuartpb:
The source for said iframe is at https://github.com/muvaster/embed.whatcardboard.com/

borismus:
A few things:

1. I'm wary of introducing an external dependency on embed.whatcardboard.com. What happens when the domain goes down?
2. The real solution is to implement a QR code scanner. QR codes aren't my favorite, but picking from a list of 1000 different Cardboard variants is strictly worse.

If you'd like to help work on the real solution (QR scanner), let's talk more!

ashconnell:
I tend to agree with @borismus

1. Right now, I get a 404 on embed.whatcardboard.com 
2. Forkers can make minor adjustments for their use cases but you can't fork a website. 

It would be slightly less scary if it was just an external web api instead of an iframe - but even then it needs to be reliable (ie load tested and balanced)

stuartpb:
I'm composing my full response to this now, but I should ponit out that the site is not down - there's no content at the https://embed.whatcardboard.com/ site root (I've had no call to make any), but the iframe, which is at https://embed.whatcardboard.com/v0/1/iframe/, loads just fine.

stuartpb:
The main advantage of implementing this as an external domain via iframe, the reason I want to do it this way, is because this is the only implementation I can think of that lets polyfilled WebVR sites *share the user's Google Cardboard configuration information*, the way native apps do, without having to reconfigure their viewer for every single site.

> I'm wary of introducing an external dependency on embed.whatcardboard.com. What happens when the domain goes down?

It's just a static GitHub Pages site going through CloudFlare - in my experience, these are two of the most reliable services a site can *be* hosted with. Besides that, the iframe is cached via appcache (with a ServiceWorker version planned), so if the user has *ever* loaded the iframe through *any* page, they'll already have it cached and won't notice the difference.

In the further unlikely case that *both* of these scenarios have failed, the script could implement some kind of fallback (indeed, right now it defaults to assuming Cardboard V2). You *could* implement a (redundant) QR code scanner, but in the scenario where a mobile device can't find embed.whatcardboard.com, the most likely case is that the network is altogether inaccessible, meaning that *any* external requests would fail - and that means attempts to read the device's profile from a QR code would fail *as well*, since profiles generated with https://wwwgc.firebaseapp.com/ have a hard dependency on the remote goo.gl URL Shortening service.

> In fact, it seems to be down right now?

As noted above, the content isn't down - there's no *other* content (such as an index for the site root), which might have been confusing, but the iframe's content is online and usable.

>The real solution is to implement a QR code scanner.

The iframe includes QR code scanning as an option, but there are a few issues with this, from experience:

1. From a pure usability standpoint, scanning the headset's QR code is not always feasible. Many headsets only come with their QR code printed separately on a piece of documentation (which doesn't get toted around with the headset): some headsets, such as the Baofeng Small Mojing, don't include a profile QR code at *all* (which is why the profile that whatcardboard makes available for that headset is a user-contributed one).
2. From a *technical* standpoint, reading QR codes is unreliable. Not only can there be problems with getting the camera to focus at arm's length (as much a UX papercut as a glitch), or in opening the camera for image capture (most mobile browsers that could handle the WebVR polyfill will do this well enough with `accept="image/*"), but merely attempting to scan a QR code from an image is often enough to cause the entire page to crash (even on my beefy Nexus 6).
3. As noted above, QR codes from profiles are network-dependent, which introduces all the problems with third-party dependencies (indeed, I just got a notification the other day from the Google Developer Dashboard saying that it thinks my account is suspicious and I need to send them  a photo of my driver's license and a banking statement to verify my identity - it would seem the API for looking up the profiles is more likely to encounter problems than the GitHub Pages site.)

> QR codes aren't my favorite, but picking from a list of 1000 different Cardboard variants is strictly worse.

There are only 59 entries in the full list, and you can narrow that down to a handful just by typing a few characters into the search box at the top of the list. It's absolutely *not* "strictly worse"; indeed, for headsets that don't have their QR code printed on them (such as the Homido Mini), the user experience is distinctly *better*.

It's okay to not want to integrate this into the main polyfill - I originally wanted this to be a separate optional drop-in component that could be included as part of [webvr-boilerplate]() but *separate* from the core WebVR Polyfill itself - but the reason I *had* to submit this as a pull request is that, in the current design of the polyfill, there's no way that it *could* be implemented without diving deep into the guts of the multiple components that this pull request touches.

Ultimately, the ability to hot-swap in third-party implementations of the VR picker is what I need from webvr-polyfill: I'd be happy to tweak this patch in any direction, so long as the end product lets me implement this approach (without having to maintain a fork).

gmarinov:
@stuartpb 

> Ultimately, the ability to hot-swap in third-party implementations of the VR picker is what I need from webvr-polyfill

sounds like, and from the code it looks like, you could simply submit a patch that allows the CardboardVRDisplay to integrate with a 3rd party ViewerSelector. This would be quite easy if the polyfill somehow made itself public. (e.g. window.webvrpolyfill) as it has some nice access methods already. (getDisplays() being one)

PR #74, opened 2016-04-06
Add approximateInverse function for inverse distortion coefficients

stuartpb:
This allows approximate inverse distortion coefficients to be calculated from a headset's specified distortion coefficients, so that headsets can be specified without having them precomputed (as they're specified in #73).

### DIFF
--- a/src/distortion/distortion.js
+++ b/src/distortion/distortion.js
@@ -28,7 +28,7 @@ Distortion.prototype.distortInverse = function(radius) {
     dr0 = dr1;
   }
   return r1;
-}
+};
 
 /**
  * Distorts a radius by its distortion factor from the center of the lenses.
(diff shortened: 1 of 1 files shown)


src/distortion/distortion.js
```diff
@@ -43,6 +43,139 @@ Distortion.prototype.distort = function(radius) {
     ret = r2 * (ret + this.coefficients[i]);
   }
   return (ret + 1) * radius;
-}
+};
+
+// Functions below roughly ported from
```
borismus: Please use /* ... */ for comment blocks.

src/distortion/distortion.js
```diff
@@ -43,6 +43,139 @@ Distortion.prototype.distort = function(radius) {
...
+//   (transpose(A) * A) * x = transpose(A) * y   <- associativity
+//   x = inverse(transpose(A) * A) * transpose(A) * y  <- solve for x
+// Matrix A's row count (first index) must match y's value count.  A's column count (second index)
+// determines the length of the result vector x.
+Distortion.prototype.solveLeastSquares_ = function(matA, vecY) {
+  var i, j, k, sum;
+  var numSamples = matA.length;
+  var numCoefficients = matA[0].length;
+  if (numSamples != vecY.Length) {
+    throw new Error("Matrix / vector dimension mismatch");
```
borismus: Please using single quotes.

borismus:
Thanks for this. Could you possibly add some unit tests? It's easy to do and would instill more confidence.

borismus:
Ping

Issue #59, opened 2016-03-29
Mobile has trouble with "forward" in landscape mode


scottgarner:
When viewing the boilerplate demo on both iOS and Android devices, the rotating cube appears in front of me when holding the device in portrait mode.

When in landscape mode, the cube appears either 90 degrees to my left or right depending on whether I'm in landscape left or landscape right.

This happens even when I manually call for a resetPose. It seems that device orientation needs to be accounted for in the pose reset.

scottgarner:
My temporary workaround for this was to apply the screen rotation Q in `resetPose`.

```
  this.resetQ.copy(this.filter.getOrientation());
  this.resetQ.multiply(this.worldToScreenQ);
```

Unfortunately, that wouldn't work on Android because of the screen locking stuff. It seems that `window.orientation` just gives the locked orientation (with no indication that it is indeed locked), so the rotation was wrong if you entered VR mode form anything other than `landscape-primary`.

My workaround there was to just disable screen locking, though then I had to mess with the instructions display to make it work with the fullscreen wrapper.

scottgarner:
Just to clarify this issue, it happens only when starting in landscape mode—"forward" appears 90 degrees to the left or right depending on device orientation.

scottgarner:
I've tinkered with this a bit more and still haven't come to a satisfactory solution.

To summarize, the initial world transformation should account for the device orientation, as should calls to resetPose. Unfortunately, when the orientation is locked on Android, it always returns the locked value, not the true orientation, which makes it difficult to account for.

I'm not sure how to fix this without knowing the "true" orientation.

zazabe:
We are using a "preparation" step at access the VR mode:
- unlock the orientation before (if it was locked)
- display some instructions, asking the user to turn his phone
- detect orientation changes and only display the button to enter VR when the phone is in landscape
- when the user click on the button, reset the device sensor + do all the VR stuff (fullscreen, orientation lock, distortion, etc...)

We have a trick to detect the orientation even if it's locked at the OS level, we use the `event.accelerationIncludingGravity.y` value on `devicemotion`, the phone should be in landscape when the value is around 0 (it works only if the phone is perpendicular to the earth actually)...

We are still using an old version of the WebVR polyfill, without the `requestPresent` method, but anyhow, those steps have to be done before entering VR, so before `requestPresent`.

borismus:
[...]

PR #85, opened 2016-04-08
Adjust for initial device rotation

scottgarner:
Adjusts `this.filterToWorld` based on initial device orientation to address issue #59;


src/sensor-fusion/fusion-pose-sensor.js
```diff
@@ -44,6 +44,11 @@ function FusionPoseSensor() {
     this.filterToWorldQ.setFromAxisAngle(new THREE.Vector3(1, 0, 0), -Math.PI/2);
   }
 
+  //Adjust for initial device orientation
```
borismus: Space after //, period after comment.

src/sensor-fusion/fusion-pose-sensor.js
```diff
@@ -44,6 +44,11 @@ function FusionPoseSensor() {
...
+  var orientationQ = new THREE.Quaternion();
+  orientationQ.setFromAxisAngle(new THREE.Vector3(0, 0, 1), window.orientation/Math.PI/2);
```
borismus: Avoid foo/bar/baz, it's really confusing.

scottgarner:
Thank you for the feedback. Unfortunately, I realized that this approach takes care of the initial orientation, but doesn't handle resets with the current `resetPose()` approach. I'm a little stuck on where to account for what.

PR #53, opened 2016-03-24
allow cardboard UI to be disabled (fixes #52)

cvan:
dependent on #51

PR #97, opened 2016-04-29
Add flags to disable rotation instructions and ui on stereo mode

dmarcos:
We're handling UX in a different way and we need to override the default one. These pieces might better placed in https://github.com/borismus/webvr-boilerplate. IMO, the webvr-polyfill should just polyfill the API and not be opinionated about UX flows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,11 +43,17 @@ are supported:
 
 ```javascript
 WebVRConfig = {
+  // Flag to disabled the UI in VR Mode.
+  CARDBOARD_UI_DISABLED: false, // Default: false
+
   // Forces availability of VR mode, even for non-mobile devices.
   FORCE_ENABLE_VR: true, // Default: false.
 
   // Complementary filter coefficient. 0 for accelerometer, 1 for gyro.
   K_FILTER: 0.98, // Default: 0.98.
+
+  // Flag to disable the instructions to rotate your device.
+  ROTATE_INSTRUCTIONS_DISABLED: false, // Default: false.
 
   // How far into the future to predict during fast motion (in seconds).
   PREDICTION_TIME_S: 0.040, // Default: 0.040.
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -1,4 +1,97 @@
 (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(_dereq_,module,exports){
+// shim for using process in browser
+
+var process = module.exports = {};
+var queue = [];
+var draining = false;
+var currentQueue;
+var queueIndex = -1;
+
+function cleanUpNextTick() {
+    draining = false;
+    if (currentQueue.length) {
+        queue = currentQueue.concat(queue);
+    } else {
+        queueIndex = -1;
+    }
+    if (queue.length) {
+        drainQueue();
+    }
+}
+
+function drainQueue() {
+    if (draining) {
+        return;
+    }
+    var timeout = setTimeout(cleanUpNextTick);
+    draining = true;
+
+    var len = queue.length;
+    while(len) {
+        currentQueue = queue;
+        queue = [];
+        while (++queueIndex < len) {
+            if (currentQueue) {
+                currentQueue[queueIndex].run();
+            }
+        }
+        queueIndex = -1;
+        len = queue.length;
+    }
+    currentQueue = null;
+    draining = false;
+    clearTimeout(timeout);
+}
+
+process.nextTick = function (fun) {
+    var args = new Array(arguments.length - 1);
+    if (arguments.length > 1) {
+        for (var i = 1; i < arguments.length; i++) {
+            args[i - 1] = arguments[i];
+        }
+    }
+    queue.push(new Item(fun, args));
+    if (queue.length === 1 && !draining) {
+        setTimeout(drainQueue, 0);
+    }
+};
+
+// v8 likes predictible objects
+function Item(fun, array) {
+    this.fun = fun;
+    this.array = array;
+}
+Item.prototype.run = function () {
+    this.fun.apply(null, this.array);
+};
+process.title = 'browser';
+process.browser = true;
+process.env = {};
+process.argv = [];
+process.version = ''; // empty string to avoid regexp issues
+process.versions = {};
+
+function noop() {}
+
+process.on = noop;
+process.addListener = noop;
+process.once = noop;
+process.off = noop;
+process.removeListener = noop;
+process.removeAllListeners = noop;
+process.emit = noop;
+
+process.binding = function (name) {
+    throw new Error('process.binding is not supported');
+};
+
+process.cwd = function () { return '/' };
+process.chdir = function (dir) {
+    throw new Error('process.chdir is not supported');
+};
+process.umask = function() { return 0; };
+
+},{}],2:[function(_dereq_,module,exports){
 (function (process,global){
 /*!
  * @overview es6-promise - a tiny implementation of Promises/A+.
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -43,12 +43,19 @@ are supported:
 
 ```javascript
 WebVRConfig = {
+
+  // Flag to disabled the UI in VR Mode
+  CARDBOARD_UI_DISABLED: false, // Default: false
```
cvan: update `src/main` too. and we can close my PR now: #53

cvan: IMO, it should be `CARDBOARD_UI_ENABLED` and just make sure it's set in `src/main`

build/webvr-polyfill.js
```diff
@@ -2405,7 +2411,9 @@ function CardboardVRDisplay() {
   // Set the correct initial viewer.
   this.deviceInfo_.setViewer(this.viewerSelector_.getCurrentViewer());
 
-  this.rotateInstructions_ = new RotateInstructions();
+  if (WebVRConfig.ROTATE_INSTRUCTIONS_DISABLED) {
```
cvan: this looks wrong

build/webvr-polyfill.js
```diff
@@ -2489,7 +2497,7 @@ CardboardVRDisplay.prototype.beginPresent_ = function() {
     this.exitPresent();
   }.bind(this));
 
-  if (Util.isLandscapeMode() && Util.isMobile()) {
+  if (this.rotateInstructions_ &&Util.isLandscapeMode() && Util.isMobile()) {
```
cvan: need space after `&&`

src/cardboard-vr-display.js
```diff
@@ -136,7 +138,7 @@ CardboardVRDisplay.prototype.beginPresent_ = function() {
     this.exitPresent();
```
cvan: should handle https://github.com/borismus/webvr-polyfill/pull/97/files#diff-55f22139cda256d0c2e29cb114cdf4f0R121 and https://github.com/borismus/webvr-polyfill/pull/97/files#diff-55f22139cda256d0c2e29cb114cdf4f0R133 too

README.md
```diff
@@ -43,12 +43,19 @@ are supported:
 
 ```javascript
 WebVRConfig = {
+
```
cvan: newline unnecessary

README.md
```diff
@@ -43,12 +43,19 @@ are supported:
...
+  // Flag to disabled the UI in VR Mode
```
cvan: I would be specific and mention the back button and the settings gear and the eye line. also should end with a period to be consistent with punctuation of sentences elsewhere.

README.md
```diff
@@ -43,12 +43,19 @@ are supported:
...
+  CARDBOARD_UI_DISABLED: false, // Default: false
+
   // Forces availability of VR mode, even for non-mobile devices.
   FORCE_ENABLE_VR: true, // Default: false.
 
   // Complementary filter coefficient. 0 for accelerometer, 1 for gyro.
   K_FILTER: 0.98, // Default: 0.98.
 
+  // Flag to disable the instructions to rotate your device
```
cvan: period.

README.md
```diff
@@ -43,12 +43,19 @@ are supported:
...
+  ROTATE_INSTRUCTIONS_DISABLED: false, // Default: false
```
cvan: would also rename this one

dmarcos: I was just following the pattern laud out by the other flags ```XXX_disabled```. They all should be consistent.

build/webvr-polyfill.js
dmarcos: Consider using a linter. Looking for syntax nits is a bad use of human resources. Let's use machines until they rebel against us and demand rights.

borismus:
The polyfill should do what a WebVR implementation might do. UX components like the Cardboard transition screen is something that a native WebVR implementation might have. I'd prefer to have fewer configuration options, so let's chat about what should and shouldn't be in scope for the polyfill.

dmarcos:
But it's something that it's not currently speced in the webvr APIs and applications will probably want to have control over. AFRAME has its own style for the "rotate your device" prompt and we need to override it. The general expectation is that a polyfill gives you a "fake" implementation of an API when it's not natively available. Anything on top should be provided by a higher level API like the boilerplate.

dmarcos:
I've modified the PR based on @cvan suggestions

cvan:
I agree that developers should not be forced to show the Orientation Rotation modal nor the Cardboard UI. Those are cosmetic interface elements that have nothing to do with the WebVR APIs. IIRC, the UI was previously available only from the webvr-boilerplate until @toji moved them over to the polyfill here. What's the difference then supposed to be from the webvr-boilerplate and webvr-polyfill?

toji:
I don't oppose adding a flag to disable this, but as an FYI I think it's very likely that the native Chrome implementation of WebVR on Android will ship with a similar transition splash screen. Assuming it fades fast enough when you put the phone in landscape orientation I can't imagine it will actually slow anyone down. You need time to physically place the phone in the harness.

borismus:
All flags should have their defaults set in https://github.com/borismus/webvr-polyfill/blob/master/src/main.js#L19, could you add CARDBOARD_UI_DISABLED & ROTATE_INSTRUCTIONS_DISABLED and their defaults there?

As for naming, I'm fine with the FOO_BAR_DISABLED for these UI tweaks. Just keep in mind that I don't think we have a WebVRConfig planned for the WebVR native implementation :)

README.md
```diff
@@ -43,12 +43,18 @@ are supported:
 
 ```javascript
 WebVRConfig = {
+  // Flag to disabled the UI in VR Mode.
+  CARDBOARD_UI_DISABLED: false, // Default: false
```
cvan: per my earlier comment, you should update `src/main.js` too

README.md
```diff
@@ -43,12 +43,18 @@ are supported:
...
+
   // Forces availability of VR mode, even for non-mobile devices.
   FORCE_ENABLE_VR: true, // Default: false.
 
   // Complementary filter coefficient. 0 for accelerometer, 1 for gyro.
   K_FILTER: 0.98, // Default: 0.98.
 
+  // Flag to disable the instructions to rotate your device.
+  ROTATE_INSTRUCTIONS_DISABLED: false, // Default: false.
```
cvan: IMO, all boolean names should be changed to be truthy ones. we already have `FORCE_ENABLE_VR`, `ENABLE_DEPRECATED_API`, `DEFER_INITIALIZATION`, `MOUSE_KEYBOARD_CONTROLS_DISABLED`. it's super inconsistent.

borismus: What do you mean by "truthy"? I'm totally open to renaming, so please make concrete proposals.

dmarcos:
I updated ```src/main.js```

dmarcos:
Let me know if you want me to still change the name of the flags. If not feel free to merge.

src/cardboard-vr-display.js
borismus: Those links don't work for me.

build/webvr-polyfill.js
```diff
@@ -2367,9 +2422,10 @@ function CardboardVRDisplay() {
   // Set the correct initial viewer.
   this.deviceInfo_.setViewer(this.viewerSelector_.getCurrentViewer());
 
-  this.rotateInstructions_ = new RotateInstructions();
-
-  this.injectPresentModeCssClass_();
+  if (!WebVRConfig.ROTATE_INSTRUCTIONS_DISABLED) {
+    this.rotateInstructions_ = new RotateInstructions();
+    this.injectPresentModeCssClass_();
```
borismus: This should happen regardless of whether or not instructions are disabled. It's a fix for fullscreen mode.

src/cardboard-vr-display.js
```diff
@@ -52,9 +52,10 @@ function CardboardVRDisplay() {
   // Set the correct initial viewer.
   this.deviceInfo_.setViewer(this.viewerSelector_.getCurrentViewer());
 
-  this.rotateInstructions_ = new RotateInstructions();
-
-  this.injectPresentModeCssClass_();
+  if (!WebVRConfig.ROTATE_INSTRUCTIONS_DISABLED) {
+    this.rotateInstructions_ = new RotateInstructions();
+    this.injectPresentModeCssClass_();
```
borismus: This should happen regardless of whether or not instructions are disabled (ie. out of the if statement). It's a fix for fullscreen mode.

dmarcos:
updated

build/webvr-polyfill.js
borismus: Please rebuild.

dmarcos: Rebuilt

README.md
cvan: filed issue #100 and opened PR #101

PR #99, opened 2016-05-06
rotate instructions interstitial emits event on show/hide

### LINKED ISSUE #93: a way for the rotate instructions interstitial to announce its show/hide-ing
e.g. `document.body.dispatchEvent(new CustomEvent('VRInstructionsShown'))`

gmarinov:
resolves #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,9 @@ to make the VR experience more intuitive, including:
 - A gear icon to select your VR viewer.
 - A back button to exit VR mode.
 - An interstitial which only appears in portrait orientation, requesting you switch
-  into landscape orientation (if [orientation lock][ol] is not available).
+  into landscape orientation (if [orientation lock][ol] is not available). The interstitial
+  emits 'vrinstructionsshown' and 'vrinstructionshidden' custom events on the window
+  when showing and hiding itself.
 
 `MouseKeyboardVRDisplay` uses mouse events to allow you to do the equivalent of
 mouselook. It also uses keyboard arrows keys to look around the scene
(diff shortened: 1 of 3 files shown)


borismus:
For consistency, this should be on window. Also, please document it in the readme.

borismus:
Still not emitted on window object.

borismus:
This might need more thought, since the WebVR implementation will also provide an interstitial. @toji, any plans to have one event for going into VR mode, and a separate event for showing that the interstitial disappeared?

toji:
I'm curious what the use case is for these events. If the display `canPresent`, you should be able to assume that any interstitial is shown when you call `requestPresent` for the first time, and hidden when the promise returned by `requestPresent` resolves (and/or `vrdisplaypresentchange` fires.)

Beyond that, though the interstitial really should be treated as an opaque part of the UA. On desktop it's not necessary, but some UAs may introduce a permission prompt that needs to be clicked through before presentation starts. In that case, again, the signal for when presentation actually begins is the promise resolution and event.

gmarinov:
@toji here are three reasons

 - when the content in the interstitial requires modification. not everybody likes a cardboard advert on that screen.
 - to address situations in CSS/layout/positioning that result in a need to switch elements behind the scenes, or more typically a need to hide/show elements that interfere with the instructions and/or/if the renderer reading the size of its container.
- in embed situations on e.g. iOS when "enter VR" breaks out of an iframe, navigating to a new URL, then the instructions are displayed automatically without any user action (in the context of the new page)

borismus:
Native WebVR API isn't emitting these events either. Closing.

PR #106, opened 2016-05-11
iOS canvas size fix

toji:
The previous code should have worked, but iOS. So this new code appears to size the canvas correctly when entering into VR presentation mode after jumping through some really miserable hoops. Because iOS.


brianchirls:
The code looks good to me, and it works in my tests. 👍 for angry Moby Dick quotes in comments.

I do think it's worth documenting what exactly is going on here, though. It looks like the following (please correct me if I'm wrong):

1. Set some inline CSS on the wrapper element
2. When presenting, back up the existing inline CSS on the canvas
3. If iOS, set dimensions of the canvas one pixel off target value. One second later, set inline CSS to the desired values. Otherwise, just set inline CSS right away.
4. When exiting presentation, restore backed up CSS if any.

I suppose this could be made slightly more efficient by memoizing both the CSS strings, but I'm not gonna nitpick too much since it's not inside a hot animation loop.

brianchirls:
Upon further review, it seems this fix interferes with the display of the rotation warning UI elements. I didn't notice it before because my own code doesn't use that. Screenshot attached.

simulator screen shot may 11 2016 12 01 26 pm

toji:
@brianchirls Your analysis of the code is correct, though why iOS requires the delayed set is beyond me. It appears that in some circumstances (that the polyfill hits consistently) the CSS style of the canvas is ignored in favor of the canvas width/height (which seems to be set in CSS pixels, not device pixels.) Setting it to something slightly off, waiting a moment, and then setting it back forces the style to re-evaluate and gets the right size. Fortunately in the default case the canvas will be hidden by the interstitial while this happens. (Fun fact: waiting only 100ms is apparently not long enough to appease the iOS style Gods.)

I had to move away from the otherwise cleaner style sheet injection method, since it's not as straightforward to do this sort of pixel twiddling with that method. Coincidentally, the current method is what @borismus wanted to do in the first place and I argued against it. Oops! Sorry Boris!

I can look into the issue with the rotation UI, but I won't have a chance to debug it till next week.

build/webvr-polyfill.js
```diff
@@ -1186,6 +1190,36 @@ VRDisplay.prototype.wrapForFullscreen = function(element) {
...
+    self.fullscreenElement_.setAttribute('style', cssProperties.join('; ') + ';');
+  }
+
+  if (Util.isIOS()) {
+    // "To the last I grapple with thee;
+    //  from hell's heart I stab at thee;
+    //  for hate's sake I spit my last breath at thee."
+    // -- Moby Dick, by Herman Melville
+    this.fullscreenElement_.setAttribute('style', 'width: ' + (Math.max(screen.width, screen.height) - 1) + 'px;');
+    setTimeout(applyFullscreenElementStyle, 1000);
```
cvan: why?

toji: Yeah, I know. I explained to Brian the best I could on the commit comments. Really it's just "because iOS"

cvan: I mean, rather, what's the cause? are the dimensions wrong when you read them?

gmarinov: @cvan , a few guesses from my experience battling this:

 - iOS 9.3+ has a scrolling bug that results in the canvas getting these wrong dimensions. zooming (css: zoom) to 1.1 and back to 1 fixed it for us.
 - iOS 7+ sometimes reports the wrong dimensions in some cases. (e.g. when the browser chrome is on screen)
 - iOS seems to issue orientationchange-related events in a different order to Android so one has to account for this.
 - Too many things end up trying to resize the canvas - especially when sized via CSS absolute positioning.

We studied this for a while, then had decent success by disabling automatic sizing, sizing things ourselves, and then passing the dimensions and pixel ratio to the renderer/effect. I think there should be an option in the config to allow this.

brianchirls:
I believe the issue with the rotation UI is the same as #111

build/webvr-polyfill.js
```diff
@@ -1186,6 +1190,36 @@ VRDisplay.prototype.wrapForFullscreen = function(element) {
...
+      'padding: 0',
+    ];
+    self.fullscreenElement_.setAttribute('style', cssProperties.join('; ') + ';');
+  }
+
+  if (Util.isIOS()) {
+    // "To the last I grapple with thee;
+    //  from hell's heart I stab at thee;
+    //  for hate's sake I spit my last breath at thee."
+    // -- Moby Dick, by Herman Melville
```
borismus: LGTM

borismus:
Ok, this is now busted if BUFFER_SCALE is not 0.5. Seems like another iOS bug, which gets resolved if you set the size of the canvas to the same thing.

PR #112, opened 2016-05-12
Fix #111 (Cardboard selector was underneath the canvas)

### LINKED ISSUE #111: Cardboard selector opens underneath the canvas and isn't visible
The canvas has a `z-index:999999` on it and the selector overlay and dialog don't.

idlewan:


src/viewer-selector.js
```diff
@@ -124,6 +125,7 @@ ViewerSelector.prototype.createDialog_ = function(options) {
   s.background = '#fafafa';
   s.fontFamily = "'Roboto', sans-serif";
   s.boxShadow = '0px 5px 20px #666';
+  s["z-index"] = 120;
```
borismus: Are you sure this works... maybe use s.zIndex?

idlewan: Yes, it works when it's written that way as well as when it's written `s.zIndex`.

idlewan:
I now changed `s["z-index"]` to `s.zIndex` for code styling consistency.

src/viewer-selector.js
cvan: can't this just be `2` (see [here])(https://github.com/borismus/webvr-polyfill/blob/a18c4af34f/examples/basic/index.html#L31)?

build/webvr-polyfill.js
```diff
@@ -2604,7 +2604,7 @@ CardboardVRDisplay.prototype.injectPresentModeCssClass_ = function() {
     'left: 0 !important',
     'right: 0 !important',
     'bottom: 0 !important',
-    'z-index: 999999 !important',
+    'z-index: 100 !important',
```
cvan: why not start the `z-index`es at `1`, `2`, `3`?

idlewan: I guess if a z-index was needed, it was to be in front of other elements on the page that might have a z-index. So 100 would be big enough to be over these other elements?

Ideally you shouldn't need to set a z-index. Also, a library shouldn't force `!important` styles, it should be overridable by the user of the library.
Maybe the z-index was required for iOS that doesn't implement the fullscreen api?

borismus:
Thanks for flagging this @idlewan, but I've fixed it as part of merging https://github.com/borismus/webvr-polyfill/pull/106.

PR #125, opened 2016-06-07
fix orientation.lock() error handling

jeromeetienne:
- if screen.orientation.lock() fails, catch the error. 

thus it doesnt trigger an exception in js console. 

im not sure it is a bug. But when i launch webvr-polyfill in chrome desktop and toggle 'device mode' in devtools, this exception is triggered everytime you goto to vr mode.

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -188,7 +188,9 @@ VRDisplay.prototype.requestPresent = function(layers) {
         self.isPresenting = (fullscreenElement === actualFullscreenElement);
         if (self.isPresenting) {
           if (screen.orientation && screen.orientation.lock) {
-            screen.orientation.lock('landscape-primary');
+            screen.orientation.lock('landscape-primary').catch(function(error){
+                    console.error('screen.orientation.lock() failed due to', error.message)
+            });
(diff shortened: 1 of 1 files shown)


src/base.js
```diff
@@ -187,7 +187,9 @@ VRDisplay.prototype.requestPresent = function(layers) {
         self.isPresenting = (fullscreenElement === actualFullscreenElement);
         if (self.isPresenting) {
           if (screen.orientation && screen.orientation.lock) {
-            screen.orientation.lock('landscape-primary');
+            screen.orientation.lock('landscape-primary').catch(function(error){
+                    console.error('screen.orientation.lock() failed due to', error.message)
```
borismus: Plz fix indentation, should be two spaces

jeromeetienne: oops, it is fixed now, i think :)

borismus: Don't see the changes.

borismus:
Thanks!

jeromeetienne:
@borismus im not a git expert, i think i did it tho :)

Issue #117, opened 2016-05-28
mixed up warning messages


blq:
Minor. 
When the method `DeviceInfo.determineDevice_` uses fallback measurements it mixes up the console warning messages between iOS and Android.

borismus:
Thanks, please submit a PR!

jeromeetienne:
i did a pull request fixing this issue in #127

jeromeetienne:
maybe we could close this bug, as the pullrequest got merged

blq:
thanks for fixing!

PR #127, opened 2016-06-09
fix mixed-up messages in device info

jeromeetienne:
- the console.warn messages were just mixed up
- this put it back in order
- related to #117


borismus:
Thanks!

PR #131, opened 2016-06-10
Adjust canvas CSS so iPhone 5 will not show URL and button bars.

### LINKED ISSUE #124: No full screen on iPhone 5
iPhone does not support real full screen mode, but if you start VR in portrait orientation and then rotate to landscape, the URL bar and various buttons disappear, approximating full screen. This works fine on iPhone 6, but not iPhone 5. Instead, the browser chrome very briefly disappears and then immediately re-appears. This did not use to happen.

There is some related discussion on this in #108.

brianchirls:
Fixes #124

Added padding on right and bottom of canvas, because iPhone 5 will not
hide the URL bar unless content is bigger than the screen.
This will not be visible as long as the container element (e.g. body)
is set to 'overflow: hidden'.

Also, iPhone 5 doesn't like 'position: fixed', so I changed it to 'absolute'


dmarcos:
I'm excited about this. I had people reporting the problem on A-Frame but I had not yet investigated it (because I don't have a small iphone around :) ). Thanks open source!

brianchirls:
Thanks for the comment, @dmarcos. I recommend getting iOS Simulator as part of xcode. I also recommend testing on iOS 8. We actually had a very wealthy client who had an iPhone 5 running iOS 8 and IE11, so it's important to test on that stuff.

That reminds me, @borismus, there's a [note](https://github.com/borismus/webvr-polyfill/blob/master/src/util.js#L193) about removing a workaround once iOS fixes a bug. Please keep in mind that people don't always upgrade right away, so we should hold off on removing workarounds for at least one old version, which for Apple could be years. :-/

src/cardboard-vr-display.js
```diff
@@ -233,7 +237,7 @@ CardboardVRDisplay.prototype.onResize_ = function(e) {
       'height: ' + Math.min(screen.height, screen.width) + 'px',
       'border: 0',
       'margin: 0',
-      'padding: 0',
+      'padding: 0 10px 10px 0',
```
cvan: should probably add a comment?

brianchirls: See lines 228 to 231

brianchirls:
Bump. @borismus do you mind taking a look at this please?

borismus:
Thanks Brian, did you test on Android & iOS? I don't have an iPhone 5 to test on.

brianchirls:
Yes, I have tested on Android and iOS. If someone else would like to confirm, that's cool. @borismus, if you have a Mac, I recommend the iOS simulator that comes with xcode. It actually emulates the hardware and iOS, so it's much more representative of a real phone than, the mobile emulation in Chrome dev tools. I do have access to a real iPhone 5, but the simulator behaves the same way for this particular issue.

lennerd:
As mentioned here https://github.com/googlevr/webvr-polyfill/pull/155#issuecomment-279971508 this will cause a bug on newer iPhones creating visible padding on the right and bottom.

Not quite sure how to fix it and make a PR for this though. Any ideas?

lennerd:
Tried around a little bit more. What do you think about adding `content-box` as box sizing method so the canvas padding is not respected when setting the size of the canvas?

cvan:
@lennerd does that work? do you have a test case or demo somewhere?

lennerd:
I created a PR with two test cases here: https://github.com/googlevr/webvr-polyfill/pull/195

PR #135, opened 2016-07-05
Support VRLayer bounds configuration

brianchirls:
- support calls to `requestPresent` when already presenting, to update layer config
- update distorter dimensions on vrdisplaypresentchange
- copy VRLayer config so it's not accidentally overwritten

There is a test example here http://output.jsbin.com/fohuqen

I have submitted a [pull request at three.js](https://github.com/mrdoob/three.js/pull/9290) to make `VREffect` correctly handle changing `VRLayer` bounds, and the example above uses that code. However, this pull request is not strictly dependent on it, so you shouldn't need to wait for the next three.js to merge.

### DIFF
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -1113,6 +1113,9 @@ var WakeLock = _dereq_('./wakelock.js');
 var nextDisplayId = 1000;
 var hasShowDeprecationWarning = false;
 
+var defaultLeftBounds = [0, 0, 0.5, 1];
+var defaultRightBounds = [0.5, 0, 0.5, 1];
+
 /**
  * The base class for all VR displays.
  */
@@ -1244,10 +1247,7 @@ VRDisplay.prototype.removeFullscreenWrapper = function() {
 };
 
 VRDisplay.prototype.requestPresent = function(layers) {
-  if (this.isPresenting) {
-    console.error('Already presenting!');
-    return;
-  }
+  var wasPresenting = this.isPresenting;
   var self = this;
 
   if (!(layers instanceof Array)) {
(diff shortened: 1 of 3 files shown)


build/webvr-polyfill.js
```diff
@@ -1269,7 +1269,51 @@ VRDisplay.prototype.requestPresent = function(layers) {
...
+          changed = true;
+        }
+        if (layer.rightBounds[i] !== rightBounds[i]) {
+          layer.rightBounds[i] = rightBounds[i];
+          changed = true;
+        }
+      }
+
+      if (changed) {
+        self.fireVRDisplayPresentChange_();
```
toji: I'm not sure this is the desired behavior (though the spec is definitely ambiguous about it.) I think we only want to fire the event when we start presenting and end presenting. Even changing sources is probably not cause to fire it.

brianchirls: This is how the browser currently behaves, so I'm copying that. Although, I don't know for sure if the browser fires the event when `requestPresent` has been called with all identical values - I just did it here to err on the side of efficiency.

But the browser definitely does currently fire the event when those values change, which is why my example is able to work at all. Even we decide not to fire `vrdisplaypresentchange`, there should be some other event to determine when the parameters change. I don't want to have to query these values on every single frame.

For reference: https://github.com/w3c/webvr/issues/23

toji:
Fair enough. LGTM for now, with some spec clarification needed later. Fortunately the apps that need to be robust to redundant `presentchange` events are the ones that are requesting them in the first place, so maybe it's a non-issue. :)

brianchirls:
Cool.

I think we all need to be encouraging code that is idempotent so it's as robust as possible. There are a lot of weird state changes going on, especially in and around the polyfill. Fortunately, most of the event listeners I've seen already check - for example - the state of `isPresenting`, because the event sometimes fires twice anyway, depending on how it's set up.

brianchirls:
Friendly bump.

borismus:
Right, sorry. Just returned from two week vacation.

brianchirls:
Oh, right. Vacation. I had to look that one up. Sounds nice. Hope you had a good time.

Thanks for merging.

PR #142, opened 2016-08-12
Add ability to remove event listeners on FusionPoseSensor.js

allanweir:
Hey, great work on the polyfill! I'm using just the pose sensor part of the project at the moment which is really useful but I noticed it carries on listening for motion events after it's destroyed so I've added a method to stop or start them again. Hopefully you find it useful, let me know your thoughts.
Thanks,
Al


borismus:
Thanks Allan! At the moment, the polyfill never stops. I guess I don't see the point of having this extra code, since it's never called unless you use it standalone as you do. If the polyfill had a way of stopping, it would make sense, but I just don't see when that would be useful.

PR #148, opened 2016-09-11
Add options to WebVRConfig for customize Cardboard viewer list

rocwind:
It supports only CardboardV1 and V2 in the polyfill library, but there are many other Cardboard viewer devices out there. To make it configurable for client application, this pull request adds following viewer setting to `WebVRConfig`:
* CARDBOARD_VIEWERS: for adding more viewer options
* DEFAULT_CARDBOARD_VIEWER: id of the default viewer

### DIFF
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -1934,7 +1934,9 @@ var DEFAULT_IOS = new Device({
 });
 
 
-var Viewers = {
+var Viewers = undefined;
+
+var DEFAULT_VIEWERS = {
   CardboardV1: new CardboardViewer({
     id: 'CardboardV1',
     label: 'Cardboard I/O 2014',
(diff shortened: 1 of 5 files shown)


build/webvr-polyfill.js
```diff
@@ -1954,17 +1953,11 @@ var Viewers = {
     interLensDistance: 0.064,
     baselineLensDistance: 0.035,
     screenLensDistance: 0.039,
-    distortionCoefficients: [0.34, 0.55],
-    inverseCoefficients: [-0.33836704, -0.18162185, 0.862655, -1.2462051,
```
rocwind: It seems `inverseCoefficients` (and `DEFAULT_LEFT(RIGHT)_CENTER`) are no longer used.

borismus:
Sorry, but the list of Cardboard viewers is too long to support in this way. The only scalable approach to this is to use QR code scanning like the native app does.

Issue #47, opened 2016-03-24
Export `WebVRPolyfill` as standalone module


cvan:
In [`main.js`](https://github.com/borismus/webvr-polyfill/blob/master/src/main.js), instead of calling `new WebVRPolyfill();` or `window.InitializeWebVRPolyfill` (which calls `new WebVRPolyfill();` at some later time), `window.WebVRPolyfill` should be exposed for browsers and for CommonJS `module.exports = WebVRPolyfill`.

This allows folks to use `window.WebVRPolyfill` or `require('webvr-polyfill')` in browsers and CommonJS, respectively.

And browserify will handle the conditional UMD (window, CommonJS, AMD) exporting for you. Just need to change `src/main.js` and have browserify export a standalone bundle.

cvan:
This isn't fixed. Why is this closed?

borismus:
I thought I had fixed it, but you're right!

jsantell:
Currently you can access the `WebVRPolyfill` object in built environments (browserify, webpack, etc.) via 

```js
// es6 import
import { WebVRPolyfill } from 'webvr-polyfill/src/webvr-polyfill';
// require
const WebVRPolyfill = require('webvr-polyfill/src/webvr-polyfill').WebVRPolyfill;
```

The use case of dropping in the script file should just seemlessly polyfill the API (which is what bundling from src/main.js does), and any custom configuration (beyond the WebVRConfig values) where you'd want to access the polyfill constructor directly can be done by accessing the constructor via require/es6 imports. Is there a common use case inbetween these two scenarios?

cvan:
I started a local branch on this. Is there interest in it? (I've been modelling it after `es6-promise`.)

jsantell:
I'd just be hesitant to do any big architectural changes without first having some solid testing, and that the script in `build/` still just auto-polyfills without further interaction, and if the imports work from https://github.com/googlevr/webvr-polyfill/issues/47#issuecomment-303794167, I'd be curious what is missing (in lieu of larger changes)

cvan:
à la [`es6-promise`](https://github.com/stefanpenner/es6-promise#auto-polyfill), this is what I'd recommend:

```js
// es6 import
import { WebVRPolyfill } from 'webvr-polyfill';
WebVRPolyfill.polyfill();

// require
const WebVRPolyfill = require('webvr-polyfill');
WebVRPolyfill.polyfill();
```

jsantell:
[...]

PR #149, opened 2016-09-12
Updated the polyfill to WebVR 1.1

toji:
Also includes functionality to shim native 1.0 implementations to the 1.1 spec.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -14,7 +14,7 @@
  */
 var Util = require('./util.js');
 var WebVRPolyfill = require('./webvr-polyfill.js').WebVRPolyfill;
-var WebVRSpecShim = require('./webvr-polyfill.js').WebVRSpecShim;
+var InstallWebVRSpecShim = require('./webvr-polyfill.js').InstallWebVRSpecShim;
 
 // Initialize a WebVRConfig just in case.
 window.WebVRConfig = Util.extend({
@@ -77,6 +77,6 @@ if (!window.WebVRConfig.DEFER_INITIALIZATION) {
   // InitializeWebVRPolyfill() will install the shim automatically when needed,
   // so this should rarely be used.
   window.InitializeSpecShim = function() {
-    new WebVRSpecShim();
+    InstallWebVRSpecShim();
   }
 }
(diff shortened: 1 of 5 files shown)


brianchirls:
The logic around `WebVRSpecShim` is a little weird and could be improved IMHO.

As currently written [here](https://github.com/borismus/webvr-polyfill/pull/149/commits/51aab8dbe4447ddf523da484180ddc19e611563b#diff-7bfb91b61bb80cff41aba297525c4dc7R38), the frameData stuff will be shimmed automatically if there's a native WebVR implementation. But if you need the polyfill, then you need to run `InitializeSpecShim`. So if you forget to run `InitializeSpecShim`, then you're stuck with a broken polyfill.

Also, there's no need for `WebVRSpecShim` to be a constructor, since it has neither methods nor properties. (On that note, I'm not sure the polyfill needs to be one either, but one thing at a time.)

So I suggest automatically polyfilling `VRFrameData` et al right inside `WebVRPolyfill` every time. There's no cost to doing so, since you're loading the JS code anyway, and it won't have any effect if it's not needed.

borismus:
This is a large change and we don't have unit tests :( 

Could you test on iOS, Android, and various desktop browsers?

toji:
Fixed the desktop issues and a few more besides, incorporated @brianchirls feedback too. Tested on Android, iOS, and Desktop Chrome/Firefox/Safari (It's installed on https://webvr.info/samples if you want to test yourself.)

src/main.js
```diff
@@ -72,4 +73,10 @@ if (!window.WebVRConfig.DEFER_INITIALIZATION) {
   window.InitializeWebVRPolyfill = function() {
     new WebVRPolyfill();
   }
+  // Call this if you want to use the shim without the rest of the polyfill.
+  // InitializeWebVRPolyfill() will install the shim automatically when needed,
+  // so this should rarely be used.
+  window.InitializeSpecShim = function() {
```
cvan: IMO, we should not be adding yet another global for the WebVR Polyfill. see #47; I have a branch of this I've done locally that moves everything to a single `WEBVR_POLYFILL`  global that is exported from `./webvr-polyfll.js` and then from here in `src/main.js`.

borismus: Agree. When would this global be called? Also why do we have the two different names `InstallWebVRSpecShim` and `InitializeSpecShim` ?

PR #155, opened 2016-09-28
Use the padding only under iOS especially if it is only for iPhone 5.

cnspaha:
Without these changes the 10px padding occurs on my Sony Xperia Z5 when starting the VR in Portrait mode and it won't go away. So just use it on iOS if it is only for iPhone 5

### DIFF
--- a/src/cardboard-vr-display.js
+++ b/src/cardboard-vr-display.js
@@ -236,6 +236,12 @@ CardboardVRDisplay.prototype.onResize_ = function(e) {
     // hide the URL bar unless content is bigger than the screen.
     // This will not be visible as long as the container element (e.g. body)
     // is set to 'overflow: hidden'.
+    var padding;
+    if (Util.isIOS()) {
+      padding = '0 10px 10px 0';
+    } else {
+      padding = '0';
+    }
     var cssProperties = [
       'position: absolute',
       'top: 0',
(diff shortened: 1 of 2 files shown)


src/cardboard-vr-display.js
```diff
@@ -236,6 +236,12 @@ CardboardVRDisplay.prototype.onResize_ = function(e) {
     // hide the URL bar unless content is bigger than the screen.
     // This will not be visible as long as the container element (e.g. body)
     // is set to 'overflow: hidden'.
+    var padding;
```
borismus: Simpler: `var padding = 0;`, and then you don't need the `else {}` clause.

PaHae: yeah u are right. changed this ;)

PaHae:
Just saw that on my iPhone 6 Plus the behaivour is the same as on the Sony Xperia Z5... also have the padding there... is there a possibility to diff between iPhone 5 and iPhone 6 so we can only use the padding on the iPhone 5?

borismus:
We can differentiate, but this seems like an overly specific hack. Could you provide a screenshot of the problem you're seeing?

cnspaha:
hmm... I would like to, but somehow now I can't manage to reproduce it... 
don't know why it works now... but if it do it is fine :D so we can forget my last comment.

borismus:
Please close then!

lennerd:
The issue still exists. I'm working with A-Frame in embedded mode and changing the orientation of the device will produce a `10px` thick border on the right and bottom. I'm using an iPhone 7.

img_4773

jamesvecore:
I'm still seeing this padding issue as well on lots of modern Android devices: Pixel, Pixel XL, 5X, 6P, etc. One key point is that this only happens when presenting to VR from portrait mode (otherwise the onResize_() function which adds the padding doesn't get called).  Here is screenshot of the issue on a nexus 5X:

screen

I added a artificial red border to highlight the issue which is the padding added at the bottom and right sides. You can also notice in the screen shot that the back and settings buttons are abnormally large.

PR #158, opened 2016-10-05
Implement navigator.vrEnabled

brianchirls:
Refer to the [spec](https://w3c.github.io/webvr/#navigator-vrenabled-attribute)

This will be necessary when running a webvr page inside an iframe that may or may not have the appropriate permissions to access webvr devices. The iframe needs to detect when it does not have permission so it can pop out into a new window or otherwise notify the user that VR is not supported.

Because of the nature of the polyfill, this behaves a little bit differently from the native implementation. Natively, `navigator.vrEnabled` requires an iframe to have the `allowvr` attribute set, but since the polyfill uses full screen instead, in this case it will require the `allowfullscreen` attribute (and/or prefixed alternatives for older browser versions).

Note: Currently, neither Firefox nor Chromium implements this property natively, but presumably they will before release, since it's in the spec.

### DIFF
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -5621,6 +5621,18 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
 
   // Provide the VRDisplay object.
   window.VRDisplay = VRDisplay;
+
+  // Provide vrEnabled
+  var that = this;
+  Object.defineProperty(navigator, 'vrEnabled', {
+    get: function () {
+      return that.isCardboardCompatible() &&
+        (document.fullscreenEnabled ||
+          document.mozFullScreenEnabled ||
+          document.webkitFullscreenEnabled ||
+          false);
+    }
+  });
 };
 
 WebVRPolyfill.prototype.enableDeprecatedPolyfill = function() {
--- a/build/webvr-polyfill.js
+++ b/build/webvr-polyfill.js
@@ -5622,11 +5622,11 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
   // Provide the VRDisplay object.
   window.VRDisplay = VRDisplay;
 
-  // Provide vrEnabled
-  var that = this;
+  // Provide navigator.vrEnabled.
+  var self = this;
   Object.defineProperty(navigator, 'vrEnabled', {
     get: function () {
-      return that.isCardboardCompatible() &&
+      return self.isCardboardCompatible() &&
         (document.fullscreenEnabled ||
           document.mozFullScreenEnabled ||
           document.webkitFullscreenEnabled ||
--- a/src/webvr-polyfill.js
+++ b/src/webvr-polyfill.js
@@ -97,10 +97,10 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
   window.VRDisplay = VRDisplay;
 
   // Provide navigator.vrEnabled.
-  var that = this;
+  var self = this;
   Object.defineProperty(navigator, 'vrEnabled', {
     get: function () {
-      return that.isCardboardCompatible() &&
+      return self.isCardboardCompatible() &&
         (document.fullscreenEnabled ||
           document.mozFullScreenEnabled ||
           document.webkitFullscreenEnabled ||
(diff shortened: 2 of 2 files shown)


src/webvr-polyfill.js
```diff
@@ -95,6 +95,18 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
 
   // Provide the VRDisplay object.
   window.VRDisplay = VRDisplay;
+
+  // Provide vrEnabled
+  var that = this;
+  Object.defineProperty(navigator, 'vrEnabled', {
+    get: function () {
+      return that.isCardboardCompatible() &&
```
cvan: perhaps just have a separate `var` for this boolean outside so you don't have to inherit the entire scope?

src/webvr-polyfill.js
```diff
@@ -95,6 +95,18 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
 
   // Provide the VRDisplay object.
   window.VRDisplay = VRDisplay;
+
+  // Provide vrEnabled
+  var that = this;
+  Object.defineProperty(navigator, 'vrEnabled', {
```
brianchirls: I thought about that. I suppose it makes sense if we can come up with a reliable way to detect it. Maybe request VR displays and return `false` if there's an error? What do you think?

Note that this needs to detect every time, since the attribute on the iframe could change at any time.

src/webvr-polyfill.js
```diff
@@ -95,6 +95,18 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
 
   // Provide the VRDisplay object.
   window.VRDisplay = VRDisplay;
+
+  // Provide vrEnabled
```
cvan: end w/ period to be consistent

cvan: and I'd say `navigator.vrEnabled`

src/webvr-polyfill.js
```diff
@@ -95,6 +95,18 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
...
+  var that = this;
+  Object.defineProperty(navigator, 'vrEnabled', {
+    get: function () {
+      return that.isCardboardCompatible() &&
+        (document.fullscreenEnabled ||
```
brianchirls: I'm not sure I'm following you. I haven't bothered to implement anything here for the deprecated API. I could do that if you think it's useful. But from the spec, I don't think `vrEnabled` requires any actual devices to be connected. I think permission on the context is sufficient, which would mean 1) having the required iframe attributes and 2) HTTPS. Full screen permission should be a valid proxy for both of those.

src/webvr-polyfill.js
cvan: disregard this above comment; I see from your other comment

src/webvr-polyfill.js
brianchirls: Fixed. I don't know why github isn't hiding these comments.

brianchirls:
For reference, here is the bugzilla ticket for implementation of `navigator.vrEnabled` in Firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1306434

src/webvr-polyfill.js
cvan: oh, sorry I'm getting confused with the old API - conflating it with Fullscreen for WebVR 1.0 `<iframe allowvr>`. our naming doesn't help heh

build/webvr-polyfill.js
```diff
@@ -5621,6 +5621,18 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
 
   // Provide the VRDisplay object.
   window.VRDisplay = VRDisplay;
+
+  // Provide vrEnabled
```
cvan: need to rebuild the source one more time

src/webvr-polyfill.js
cvan: this all sounds good to me now

src/webvr-polyfill.js
```diff
@@ -95,6 +95,18 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
 
   // Provide the VRDisplay object.
   window.VRDisplay = VRDisplay;
+
+  // Provide navigator.vrEnabled.
+  var that = this;
```
cvan: nit: `self` is used [elsewhere](https://github.com/brianchirls/webvr-polyfill/blob/28d66968/src/webvr-polyfill.js#L135)

borismus:
LGTM. Thanks!

PR #161, opened 2016-10-07
Added possibility to disable wakelock

cnspaha:


borismus:
Why would you want to disable wakelock?

cnspaha:
The problem I have is on iOS. 
It uses `window.location = window.location; window.stop()`
and my problem is, that sometimes my Cordova SPA is getting reloaded due to this code, seems like the window.location is sometimes faster than the window.stop, dunno...
and this is absolutely not acceptable for UX, when u are in VR and suddenly the app is getting reloaded.... So I want to disable this method for wakelock and use the cordova-plugin insomnia for native wakelock, or better I already do in our SPA.
you understand my point?

borismus:
I think Cordova usage is pretty exotic. I don't think we should add this
extra flag, sorry.

cnspaha:
I get your point, that the usage of Cordova paired with WebVR might be a bit exotic, but even then it wouldn't hurt to add a flag to disable the wakelock? Of course the default should be that the wakelock already implemented in the polyfill should be used.

build/webvr-polyfill.js
```diff
@@ -140,7 +140,10 @@ function VRDisplay() {
   this.fullscreenChangeHandler_ = null;
   this.fullscreenErrorHandler_ = null;
 
-  this.wakelock_ = new WakeLock();
+  if (!WebVRConfig.DISABLE_WAKELOCK) {
+    var WakeLock = _dereq_('./wakelock.js');
```
borismus: Also not at all a fan of this.

PR #170, opened 2016-11-01
Fixed error in strict mode

MrSlide:
The function `mat4_invert` was being assigned to a variable that was not declared, causing errors when in strict mode. It caused me problems when importing the module with Rollup.


MrSlide:
I signed it!

AlbertoElias:
I noticed the latest release didn't actually include this commit, but the last commit did bring in this change. Would it be possible to make a new release?

PR #171, opened 2016-11-02
Save and restore vertex attributes in CardboardDistorter

daredevildave:
Fixes bugs in PlayCanvas where we have cached vertex attribute state. FAO @toji ?

### DIFF
--- a/src/cardboard-distorter.js
+++ b/src/cardboard-distorter.js
@@ -43,6 +43,16 @@ var distortionFS = [
   '}',
 ].join('\n');
 
+// Vertex attribute state
+var va = 0;
+var vaBuff = [null, null];
+var vaEnabled = [null, null];
+var vaOffset = [null, null];
+var vaSize = [null, null];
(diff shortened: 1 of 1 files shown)


toji:
Logic looks good, but I've found that querying attributes like this has been one of the biggest sources of slowdown in previous iterations of the polyfill. The preferred method (ugly though it may be) would be to swap out the relevant calls on the context with ones that cache the incoming values before sending them down to the real function. See `CardboardDistorter.prototype.patch`/`CardboardDistorter.prototype.unpatch` for multiple examples.

borismus:
@daredevildave Any updates here?

daredevildave:
Sorry, haven't had a chance to update the code to the more performant version. However, we're using it with this change merged currently on PlayCanvas. I'll try and update this week.

borismus:
Ping!

borismus:
Closing for inactivity.

Issue #177, opened 2016-11-18
Polyfill for browser with Native API support but that return no VRDisplays


toji:
As we get closer to having the native WebVR API exposed in some browser there's a possibility that in some cases the native API won't provide any VRDisplays but the polyfill could still function reasonably. Currently in these scenarios having a "WebVR enabled browser" would actually appear to the user as a detriment, because the page would stop providing any WebVR capabilities where before it was able to use the Polyfill at least.

I propose that we refactor the polyfill so that rather than feature checking for the existence of the API and exiting early if it's available it instead patches the native `getVRDisplays` call with a wrapper that checks to see if the promise resolves with an empty array. If so, it supplies a polyfilled VRDisplay instead that functions the same as in the current builds when the native API isn't present at all.

This will probably be something of a short-term concern (hopefully on the order of months), but I feel the reduction in user confusion it would provide during the API's rollout is worthwhile. Any thoughts, objections?

marciot:
Why would it be a "short-term" concern? The WebVR polyfill currently provides two useful benefits. One, it polyfills the WebVR API for browsers that lacks it; and two it provides a convenient virtual VRDisplay for both mobile devices and desktops that lack a VR headset. The first benefit will diminish as more browsers implement the WebVR API, but the second benefit will be indefinite as there will always be many users who lack VR hardware.

The polyfill is providing a great service by not requiring casual game developers to handle the situations in which no native VRDisplay is available and that benefit will still be relevant once the WebVR API is well supported.

RikHeijdens:
This was one of the reasons why we stopped using the `MouseKeyboardVRDisplay`. I would totally support this.

marciot:
For my purposes, it would be sufficient if the MouseKeyboardVRDisplay were made accessible outside the polyfill, so I could write code like this:

var vrDisplay = null;
navigator.getVRDisplays().then(
   function(displays) {
      if(displays.length > 0) {
         vrDisplay = displays[0];
      } else {
         vrDisplay = new MouseKeyboardVRDisplay();
      }
   }
}

Before transitioning my game over to use the WebVR API, I had my own mouse handler but I stripped all that out once I saw that I could get orientation data from the VRDisplay. This was before I realized this what made this work on the desktop was the polyfill and that it wasn't part of the native API.

[...]

PR #180, opened 2016-11-27
Added two new modes to the polyfill

toji:
Added modes that allow the polyfill to continue working even when the
native API is present. Dev’s can choose “NO_NATIVE_API” for the
existing behavior (this is the default), “NO_NATIVE_DISPLAY” to have
the polyfill kick in when the native API is present but doesn’t provide
a display, and “ALWAYS” to always have a polyfilled VRDisplay appended
to the returned list.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -15,6 +15,19 @@
 var Util = require('./util.js');
 var WebVRPolyfill = require('./webvr-polyfill.js').WebVRPolyfill;
 
+window.WebVRPolyfillMode = {
+  // Provide a polyfilled VRDisplay only when the native API is missing.
+  // This is the default mode.
+  NO_NATIVE_API: 0,
+  // Provide a polyfilled VRDisplay if the native API is missing or is present
+  // but does not provide a VRDisplay.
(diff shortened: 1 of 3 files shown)


toji:
Fixes #177

src/main.js
```diff
@@ -15,6 +15,19 @@
 var Util = require('./util.js');
 var WebVRPolyfill = require('./webvr-polyfill.js').WebVRPolyfill;
 
+window.WebVRPolyfillMode = {
```
borismus: Do we need this distinction? Why are the non-zero modes required?

Also, I don't think this needs to be global. Looks like you expect configuration to be set via WebVRConfig, so this variable can be internal to the polyfill.

toji:
Okay, feedback addressed. Setting has been simplified to a boolean: `ALWAYS_APPEND_POLYFILL_DISPLAY`. When false the polyfill will provide a VRDisplay if the native API is either not present or present but returns no displays. When true a polyfilled display will always be appended to the returned list of VRDisplays, even when it contains a native display.

PR #185, opened 2017-01-06
Add Note 5 user agent variant to Dpdb cache

johnmaf:
We found that for one of our devices (a Samsung Note 5) the distortions were not being applied correctly, the reason being that the model number is "SM-N920W8" for Canadian devices.


johnmaf:
I signed it!!

borismus:
I'll accept this change for now, but keep in mind that DPDB is autogenerated from an internal database so at some point in the future, your change may be reverted.

PR #187, opened 2017-01-10
Add uglifyjs to dependencies

MathiasPaumgarten:
In order to run the `min` npm script one has to have `uglifyjs` on their machine. Ideally it is part of the dependencies like the other tools used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
   "devDependencies": {
     "browserify": "latest",
     "derequire": "latest",
-    "watchify": "latest"
+    "watchify": "latest",
+    "uglifyjs": "latest"
   },
   "main": "build/webvr-polyfill.js",
   "keywords": [


MathiasPaumgarten:
I signed it!

cvan:
Since [`uglifyjs`](https://github.com/mishoo/UglifyJS) hasn't been maintained in nearly 3 years, I'd recommend going with [Google Closure Compiler](https://github.com/google/closure-compiler) instead. See sample usage [here](https://github.com/mrdoob/three.js/blob/6d096f8/package.json#L33). Closure Compiler performs best in [benchmarks](https://github.com/samccone/The-cost-of-transpiling-es2015-in-2016#raw-data) too.

MathiasPaumgarten:
Makes sense. But the other thing to consider is: UglifyJS is already being used and I could see a situation where, as WebVR rolls out in more browsers, this project might not get many more updates. Leaving it at Uglify for the time being might be enough.

borismus:
We can move over to another obfuscator down the line, but since uglify is already being used in the scripts, accepting this change.

PR #195, opened 2017-02-16
Use content-box as box sizing method.

lennerd:
It's common practice to apply `box-sizing: border-box;` to elements via the `*` operator. This causes a bug when using the WebVR polyfill and the fullscreen Hack for iOS devices.

The polyfill applies some padding to overflow the content in every case causing a fullscreen without navigation and action bar. The `border-box` model adds this padding to the width and height of the canvas, causing the canvas to shift to the top left corner.

img_4775

Master: http://lennerd.taurus.uberspace.de/temp/webvr-polyfill-example/master.html
PR: http://lennerd.taurus.uberspace.de/temp/webvr-polyfill-example/pr.html

*I changed the body background to be red to see bug properly.*

This PRs simply adds `box-sizing: content-box` to the canvas when the CSS for the fullscreen view is applied.

Also see: https://github.com/googlevr/webvr-polyfill/pull/131#issuecomment-279972199


lennerd:
I signed it!

borismus:
Sorry, your PR has a bunch of DPDB changes as well. Please revert them. Also, have you tested the box-sizing changes on Android, iOS, desktop, vive, etc?

lennerd:
DPDB? Sorry, not sure, what you mean. Probably the stuff from the build file? Thought we need to add also a ready build file. Of course I can revert this.

> Also, have you tested the box-sizing changes on Android, iOS, desktop, vive, etc?

I tested this PR only on iOS. Not sure how other devices will differ though. As `content-box` is also the initial/default value for `box-sizing`, this PR just makes sure that these default behaviour is applied even when a global `box-sizing` is set, no?

borismus:
Could you please test box sizing on desktop and android? Thanks

cvan:
per @borismus' comment: are you positive this doesn't cause regressions on any browser/platform combinations?

lennerd:
I don't have an Android phone at hand at the moment, but I already tested this on a Desktop device though.

But again: my changes simply restore the default `content-box` for `box-sizing` a browser is using anyway if no own `box-sizing` is defined via a concrete or an asterik selector. I don't see why this should break any working setup. Am I missing something?

lennerd:
And please explain to me DPDB. Not able to find anything on the internet explaining it.

toji:
Fixed by #212

PR #196, opened 2017-02-24
Updated dpdb (adding Pixel)

trakout:
Pixel XL is in the database, however Pixel was not covered, and was logging "No DPDB device match".

Also I would like to recommend that https://storage.googleapis.com/cardboard-dpdb/dpdb.json be updated -- as of this pull request it is at least 2 versions behind (let me know if there should be a separate issue for this).


borismus:
I'll accept this change for now, but keep in mind that DPDB is autogenerated from an internal database so at some point in the future, your change may be reverted.

PR #204, opened 2017-03-17
Polyfill VRFrameData enabled on all devices

tbalouet:
This follow the PR from @brianchirls https://github.com/googlevr/webvr-polyfill/pull/197
By setting the condition 
`if (this.nativeWebVRAvailable && this.isCardboardCompatible() && window.VRFrameData) {`
to enable to polyfill the VRFrameData, it made it inaccessible to desktops and thus MouseKeyboard polyfill

I removed the `this.isCardboardCompatible()` condition, tested it with origin-trial WebVR on desktop/mobile not WebVR/mobile with WebVR, it all seems to work.


brianchirls:
I'm still not clear what the policy for this repo is regarding checking in builds, but otherwise this looks good to me.

The only other thought is that we might want to skip the polyfill if we're both not cardboard compatible and the mouse polyfill is disabled. The reason I put that check there in the first place was to try to avoid the extra overhead when we know we don't need it.

tbalouet:
Yes but what about people disabling mouse polyfill because they have their own system? I still think VRFrameData shouldn't be read only if there's some kind of polyfill needed

brianchirls:
@tbalouet Yeah, that's more or less what I'm suggesting. But you can never know for certain that there's no polyfill needed because the devices are available asynchronously. If you wait until the device becomes available to polyfill `VRFrameData`, by that point it will be too late and your code will break. So you have to polyfill it preemptively just in case.

src/webvr-polyfill.js
```diff
@@ -100,7 +100,7 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
   navigator.getVRDisplays = this.getVRDisplays.bind(this);
 
   // Polyfill native VRDisplay.getFrameData
-  if (this.nativeWebVRAvailable && this.isCardboardCompatible() && window.VRFrameData) {
+  if (this.nativeWebVRAvailable && window.VRFrameData) {
```
cvan: I think there should be a check for `if (window.VRDisplay.prototype.getFrameData) { … }` [here](https://github.com/tbalouet/webvr-polyfill/blob/1ac010e/src/webvr-polyfill.js#L251)

cvan:
> I'm still not clear what the policy for this repo is regarding checking in builds, but otherwise this looks good to me.

I think running `npm run build && npm run min` is acceptable. if not done in a PR, @borismus or someone else will run it in a follow-up commit.

src/webvr-polyfill.js
cvan: ping, @tbalouet and @bchirls

cvan: friendly ping

cvan:
could you also pull the latest changes on the upstream origin `master` and resolve the merge conflicts? thanks

src/webvr-polyfill.js
brianchirls: Sorry, I missed this...

Why do you think we need that check? The code is already inside a conditional for `!('VRFrameData' in window)`. Is there a case where `VRFrameData` will not exist but `getFrameData` will?

brianchirls:
@cvan That's exactly why I'm not so sure checking in builds is a good idea ;-)

cvan:
@brianchirls agreed. filed issue #216.

tbalouet:
Sorry, was postponing, rebuild done. Or should I have let the build be done when the merging to master branch will be done?

cvan:
@toji, @HalfdanJ: when you get a chance, could you review this PR?

toji:
LGTM, but we need to make sure the CLA requirements are satisfied before we can merge.

src/webvr-polyfill.js
cvan: nope, I think not, I think you're good

cvan:
thanks, @toji!

@tbalouet just see https://github.com/googlevr/webvr-polyfill/pull/204#issuecomment-287310841

tbalouet:
Ok sorry, didn't see it before, CLA is signed! Thanks!

tbalouet:
Ok guys, sorry for the mess, I guess I've commited the last fork with a corrupted github account, there's no email linked to it apparently. Should I try to fork it elsewhere?

toji:
I think it's pretty obvious who the commit is coming from and that you have, inf fact, signed the CLA, so we're good. Merged.

tbalouet:
great, thanks Brandon!

PR #206, opened 2017-03-17
Wake lock not working on Android, adapt NoSleep.js video setup

sttz:
The wake lock in webvr-polyfill was not working properly on my Moto Z Play with Chrome 56.

Attaching the debugger, I got this error which makes it sound like Chrome is having issues with the video:
«Uncaught (in promise) DOMException: Failed to load because no supported source was found.»

I then tried [NoSleep.js](https://github.com/richtr/NoSleep.js) and that was working fine. This pull request uses the video setup from NoSleep.js (adding source HTML elements, MP4+WebM videos, looping instead of ended event), with which the wake lock in webvr-polyfill is working for me again.

(I did try just using the MP4 video from NoSleep.js but that gave the same error. Not sure what the issue is but the wake lock only works for me with the WebM version).


sttz:
I signed it!

src/wakelock.js
```diff
@@ -22,22 +22,26 @@ var Util = require('./util.js');
  */
 function AndroidWakeLock() {
   var video = document.createElement('video');
+  video.setAttribute("loop", "");
```
cvan: nit: use single quotation marks

src/wakelock.js
```diff
@@ -22,22 +22,26 @@ var Util = require('./util.js');
...
 
-  video.addEventListener('ended', function() {
-    video.play();
-  });
+  function addSourceToVideo(element, type, dataURI) {
+    var source = document.createElement('source');
+    source.src = dataURI;
+    source.type = "video/" + type;
```
cvan: nit: use single quotation marks

src/wakelock.js
```diff
@@ -22,22 +22,26 @@ var Util = require('./util.js');
  */
 function AndroidWakeLock() {
   var video = document.createElement('video');
+  video.setAttribute("loop", "");
 
-  video.addEventListener('ended', function() {
-    video.play();
-  });
+  function addSourceToVideo(element, type, dataURI) {
```
cvan: nit: add a space before the parentheses, `(…)`

sttz: There are no spaces before function parentheses anywhere else in the file so I'm not sure this would be consistent?

cvan: that's fine; you can keep

toji:
Thanks for the fix, and apologies for how long it's taken to merge. We're working through a maintainers re-org.

PR #207, opened 2017-03-18
add node require regression test

kellyselden:
In A-Frame, it is now possible to require aframe in node. This is useful for certain scenarios, such as isomorphic apps.

I would like to add a regression test that preserves the ability to require webvr-polyfill. We have something similar [here](https://github.com/aframevr/aframe/blob/master/tests/node/test.js#L32-L34).

cc @dmarcos @ngokevin

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,9 +13,12 @@
   },
   "devDependencies": {
     "browserify": "latest",
+    "chai": "^3.5.0",
     "derequire": "latest",
-    "watchify": "latest",
-    "uglifyjs": "latest"
+    "jsdom": "^9.12.0",
+    "mocha": "^3.2.0",
+    "uglifyjs": "latest",
+    "watchify": "latest"
   },
   "main": "build/webvr-polyfill.js",
   "keywords": [
@@ -26,7 +29,8 @@
   "scripts": {
     "watch": "watchify src/main.js --standalone WebVRPolyfill -v -d -o 'derequire > build/webvr-polyfill.js'",
     "build": "browserify src/main.js --standalone WebVRPolyfill | derequire > build/webvr-polyfill.js",
-    "min": "browserify src/main.js --standalone WebVRPolyfill | derequire | uglifyjs -c > build/webvr-polyfill.min.js"
+    "min": "browserify src/main.js --standalone WebVRPolyfill | derequire | uglifyjs -c > build/webvr-polyfill.min.js",
+    "test": "mocha"
   },
   "repository": "googlevr/webvr-polyfill",
   "bugs": {
--- a/test/index.js
+++ b/test/index.js
@@ -1,0 +1,33 @@
+const path = require('path');
+const expect = require('chai').expect;
+const jsdom = require('jsdom');
+
+describe('node acceptance tests', function() {
+  let _window;
+
+  beforeEach(function() {
+    _window = global.window = jsdom.jsdom().defaultView;
+    global.navigator = _window.navigator;
+    global.document = _window.document;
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -13,9 +13,12 @@
   },
   "devDependencies": {
     "browserify": "latest",
+    "chai": "^3.5.0",
     "derequire": "latest",
-    "watchify": "latest",
-    "uglifyjs": "latest"
+    "jsdom": "^9.12.0",
+    "mocha": "^3.2.0",
+    "uglifyjs": "latest",
```
cvan: I'd recommend [`babili`](https://github.com/babel/babili) and others (`uglifyjs` is quite outdated and unmaintained)

package.json
```diff
@@ -13,9 +13,12 @@
...
+    "watchify": "latest"
```
cvan: I'd pin this to a specific version

test/index.js
```diff
@@ -0,0 +1,33 @@
+const path = require('path');
+const expect = require('chai').expect;
+const jsdom = require('jsdom');
+
+describe('node acceptance tests', function() {
```
cvan: nit: I'd put a space before the parentheses: `()`

test/index.js
```diff
@@ -0,0 +1,33 @@
+const path = require('path');
+const expect = require('chai').expect;
+const jsdom = require('jsdom');
+
+describe('node acceptance tests', function() {
+  let _window;
+
+  beforeEach(function() {
```
cvan: same

test/index.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  afterEach(function() {
+    delete global.window;
+    delete global.document;
+    delete global.WebVRConfig;
+  });
+
+  it('can run in node', function() {
+    require(path.join(process.cwd(), 'src/main'));
```
cvan: you can do this instead so it works on Windows:

```js
    require(path.join(process.cwd(), 'src', 'main'));
```

test/index.js
kellyselden: I don't believe google prescribes to this style:
https://github.com/googlevr/webvr-polyfill/blob/master/.jscsrc#L2
https://github.com/jscs-dev/node-jscs/blob/master/presets/google.json#L54-L56

package.json
kellyselden: The deps were out of order and npm reordered them. I didn't touch uglifyjs or watchify.

test/index.js
kellyselden: I wrote this on Windows, so it should work. That said, I'll still update it to be more correct.

test/index.js
cvan: that's fine

test/index.js
```diff
@@ -0,0 +1,33 @@
...
+        return global.WebVRConfig;
+      },
+      set(WebVRConfig) {
+        global.WebVRConfig = WebVRConfig;
+      }
+    });
+  });
+
+  afterEach(function() {
+    delete global.window;
```
cvan: friendly ping, @kellyselden

kellyselden: `_window` is a local variable only in this test file. It is not applied to `global`.

cvan: okay, sounds good - thanks

cvan:
@toji can you take a look?

cvan:
@toji, @HalfdanJ: when you get a chance, could you review this PR?

toji:
LGTM.

PR #213, opened 2017-03-31
Fix for cardboard UI sizes when entering VR from portrait mode

jamesvecore:
This fix slightly changes the way dps for the cardboard UI sizing is calculated for Android and iOS devices. When entering from portrait mode the screen.width will be the smaller dimension and this was throwing off the calculation. The code that was previously in place caused very large UI elements when entering from portrait on android devices. With this change, the same code works correctly on both iOS and Android devices. I also added some comments to explain the calculation a bit.

I've testing this a the following physical devices:

Pixel, Pixel XL, Nexus 5/6/5X/6P, Samsung S6, iPhone 5/6/7

as well as the iOS simulator for various devices.

### DIFF
--- a/src/cardboard-ui.js
+++ b/src/cardboard-ui.js
@@ -150,13 +150,12 @@ CardboardUI.prototype.onResize = function() {
 
     var midline = gl.drawingBufferWidth / 2;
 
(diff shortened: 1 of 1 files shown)


toji:
Thank you!

PR #215, opened 2017-04-07
Avoid touchstart collision

machenmusik:
Avoid touchstart collision (e.g. when no touches) by checking for touches before checking length


src/base.js
```diff
@@ -313,6 +313,13 @@ VRDisplay.prototype.exitPresent = function() {
         self.fireVRDisplayPresentChange_();
       }
 
+      if(Util.isWebViewAndroid()) {
```
cvan: what about checking `!Util.exitFullscreen()`?

src/dpdb/dpdb.js
```diff
@@ -19,7 +19,7 @@ var DPDB_CACHE = require('./dpdb-cache.js');
 var Util = require('../util.js');
 
 // Online DPDB URL.
-var ONLINE_DPDB_URL = 'https://storage.googleapis.com/cardboard-dpdb/dpdb.json';
+var ONLINE_DPDB_URL = require('./dpdb.json');
```
cvan: IMO, this should be done in a separate PR

src/touch-panner.js
```diff
@@ -45,7 +45,7 @@ TouchPanner.prototype.resetSensor = function() {
 
 TouchPanner.prototype.onTouchStart_ = function(e) {
   // Only respond if there is exactly one touch.
-  if (e.touches.length != 1) {
+  if (!e.touches || e.touches.length != 1) {
```
cvan: I know it was already like this, but can you change `!=` to be `!==`?

src/util.js
```diff
@@ -39,6 +39,15 @@ Util.isIOS = (function() {
   };
 })();
 
+Util.isWebViewAndroid = (function() {
+  var isWebViewAndroid = navigator.userAgent.indexOf('Version') !== -1 &&
+      navigator.userAgent.indexOf('Android') !== -1 &&
+      navigator.userAgent.indexOf('Chrome') !== -1;
```
cvan: this would fail for a page embedded as [a Chromium-based WebView]( https://github.com/uupaa/UserAgent.js/blob/5d90111/test/testcase.js#L178-L190), as well as [these several cases](https://github.com/uupaa/UserAgent.js/blob/5d90111/test/testcase.js#L403-L451).

src/util.js
```diff
@@ -83,6 +92,9 @@ Util.getScreenHeight = function() {
 };
 
 Util.requestFullscreen = function(element) {
+  if (Util.isWebViewAndroid()) {
```
cvan: why's this necessary? also, this code path shouldn't be accessed from iOS, right? so, we should probably avoid the initial calls to `requestFullscreen` when `!Util.isWebViewAndroid()` anyway, right?

src/util.js
```diff
@@ -99,6 +111,7 @@ Util.requestFullscreen = function(element) {
 };
 
 Util.exitFullscreen = function() {
+
```
cvan: can remove

cvan:
thanks for contributing this PR! can you add more context and a test case (and, if possible, ideally a filed issue as well)?

machenmusik:
sorry, it looks like there were other aframe fork changes pulled in for this very simple change, will try to cherry-pick just the one later unless @dmarcos or other can answer questions on the other fork changes

cvan:
@machenmusik - the changes are being looked at by @toji now, so if there's anything that's needed for A-Frame that you don't see open or merged Pull Requests for, please do open new PRs, and we'll make sure those get merged in this main, upstream repository.

cvan:
ping, @machenmusik - you good with this? need any help?

src/util.js
cvan: here's how Oculus Carmel does UA detection: https://github.com/OculusVR/ovrui/blob/master/src/Player/Player.js#L17-L19

cvan:
@machenmusik: when you get a minute, could you update this PR? we can split some of the code out into separate PRs; let me know how I can help. (just want to make sure everything in the [@aframevr fork](https://github.com/googlevr/webvr-polyfill/compare/master...aframevr:master) is up to date here in this upstream repo.)

machenmusik:
something must have changed since this was originally working, per discussion on aframevr/aframe#2669 it won't fix now, so closing

PR #224, opened 2017-05-02
Make sure ViewerSelector.selectedKey always has a default value

dhritzkiv:
Before, `this.selectedKey` would be `undefined` if `localStorage.getItem()` threw an error (such as if the user has disabled localStorage, is in Private mode, etc.).

This commit assigns the `DEFAULT_VIEWER` value in the cases of `localStorage.getItem()` returning `undefined` AND throwing an Error.

---

Without this, the viewer is also left `undefined`, and everything depending on it, such as creating new distortions (`this.distortion = new Distortion(this.viewer.distortionCoefficients)`) breaks. This has been the cause of issues for a few dozen users over the last few months.

### DIFF
--- a/src/viewer-selector.js
+++ b/src/viewer-selector.js
@@ -27,13 +27,18 @@ var CLASS_NAME = 'webvr-polyfill-viewer-selector';
  * saving the currently selected index in localStorage.
  */
 function ViewerSelector() {
-  // Try to load the selected key from local storage. If none exists, use the
-  // default key.
+  // Try to load the selected key from local storage.
   try {
-    this.selectedKey = localStorage.getItem(VIEWER_KEY) || DEFAULT_VIEWER;
+    this.selectedKey = localStorage.getItem(VIEWER_KEY);
   } catch (error) {
     console.error('Failed to load viewer profile: %s', error);
   }
+  
+  //If none exists, or if localstorage is unavailable, use the default key.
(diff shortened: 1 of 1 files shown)


dhritzkiv:
I signed it!

src/viewer-selector.js
```diff
@@ -27,13 +27,18 @@ var CLASS_NAME = 'webvr-polyfill-viewer-selector';
...
-  // default key.
+  // Try to load the selected key from local storage.
   try {
-    this.selectedKey = localStorage.getItem(VIEWER_KEY) || DEFAULT_VIEWER;
+    this.selectedKey = localStorage.getItem(VIEWER_KEY);
   } catch (error) {
     console.error('Failed to load viewer profile: %s', error);
   }
+  
+  //If none exists, or if localstorage is unavailable, use the default key.
```
cvan: good call; and [this too is handled](https://github.com/googlevr/webvr-polyfill/blob/c29bcbe42323bdd9411809b8272d9a942c4033e6/src/viewer-selector.js#L88) fine?

dhritzkiv: Yeah, that's still the best way to handle that. It doesn't break anything; it simply won't persist the user's choice to localStorage.

jsantell:
Looks good -- thanks for the PR @dhritzkiv, and thanks for reviewing @cvan!

PR #228, opened 2017-05-14
Check for touches property before deep referencing

gordonbrander:
Fixes TypeError when touching Daydream trackpad caused by the fact
that Daydream controller creates an event object without a `touches`
property.

See https://github.com/aframevr/aframe/issues/2669 for background info.


gordonbrander:
Closed in favor of https://github.com/googlevr/webvr-polyfill/pull/230

PR #229, opened 2017-05-14
avoid errors if e.touches is absent

machenmusik:
redone fix per discussion on googlevr/webvr-polyfill#215, aframevr/webvr-polyfill#4 and aframevr/aframe#2669, to isolate change for PRs to both aframevr and googlevr


machenmusik:
closing in favor of #228

PR #230, opened 2017-05-14
Fix for Daydream controller touchstart: check for touches property before deep referencing

gordonbrander:
Fixes TypeError when touching Daydream trackpad caused by the fact
that Daydream controller creates an event object without a `touches`
property.

See https://github.com/aframevr/aframe/issues/2669 for background info.

This commit was based on https://github.com/googlevr/webvr-polyfill/commit/4787fe5546e33bb82b9a3981245f39d9e1e136ba
for a cleaner merge with https://github.com/aframevr/webvr-polyfill.

### DIFF
--- a/src/touch-panner.js
+++ b/src/touch-panner.js
@@ -45,7 +45,9 @@ TouchPanner.prototype.resetSensor = function() {
 
 TouchPanner.prototype.onTouchStart_ = function(e) {
   // Only respond if there is exactly one touch.
-  if (e.touches.length != 1) {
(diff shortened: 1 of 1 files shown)


cvan:
@caseyyee this work for you? (I misplaced my Daydream controller - again!)

jsantell:
Dug into the history in aframevr/aframe#2669 and trying it out myself -- I'm not able to recreate it with my Daydream controller (maybe the update last week or so is related?), but if others are having this issue, this is still a good, safe check to have. Thanks @gordonbrander!

PR #236, opened 2017-05-25
eliminate the need to intercept WebVRConfig

kellyselden:
when running the polyfill in node

Will make it easier https://github.com/aframevr/aframe/blob/cafbadc493701ebc5da33e531ea490e20a7eaa88/tests/node/test.js#L14

and https://github.com/kellyselden/ember-aframe/blob/13a67e3d29d7af6f5d01f20ed1e5722096d38f6b/index.js#L19


jsantell:
I think there's lots of work to be done in relying less on globals throughout the code, and this is a good start to the process, same as #238. Thanks!

PR #237, opened 2017-05-25
bring back node 4 regression testing

kellyselden:
At least one project consuming this in node supports all maintained node versions. Node 4 is still maintained https://github.com/nodejs/LTS#lts-schedule1. @jsantell could I convince you to keep regression testing node 4, so that any consumers don't unintentionally break?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "node"
   - "6"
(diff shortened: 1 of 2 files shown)


jsantell:
Ah, that's an easy fix, we should definitely support it atleast until LTS is up. Thanks for sending this code in!

Issue #120, opened 2016-06-02
Use transform to remove hardcode data URIs for wakelock video, rotate-instructions icons


jeromeetienne:
- [src/rotate-instructions.js](https://github.com/borismus/webvr-polyfill/blob/master/src/rotate-instructions.js) contains a line of 53000 characters [this one](https://github.com/borismus/webvr-polyfill/blob/master/src/rotate-instructions.js#L141)
- it seems to be good at slowing down editors like sublime text, or even crashing atom 

i would not call that a bug. just wanted to share the experience :)

cvan:
I was working on a local branch that used a browserify-transform to inject the data URIs used so they wouldn't have to be embedded like so: https://github.com/borismus/webvr-polyfill/blob/a9ea232/src/rotate-instructions.js#L141

IMO, I would change the issue to be `Use transform to remove hardcode data URIs for wakelock video, rotate-instructions icons` or something.

jeromeetienne:
glad you are working on it.

change done btw :)

cvan:
would reopen

borismus:
Not familiar with browserify-transform, but open to using it instead of having the wakelock video hardcoded.

brianchirls:
As it is, this file is the single biggest script component in my entire library. Just the base64-encoded SVG is around 52k. It would be smaller (~40k) just included as text, and smaller still if you run it through [svgo](https://github.com/svg/svgo) first (~24k).

cvan:
I know it's not best practice for self-contained libraries, but how about we just split the SVG into a separate file. Let people include them when they download their copies of `webvr-polyfill`. Or let them require it conditionally, something like: `require('webvr-polyfill.assets/rotate-instructions.svg')` (I know with Webpack the syntax would be different, but with Browserify, you could do something like that). Also, just serving them on [this project's GitHub Pages](https://googlevr.github.io/webvr-polyfill/) would be helpful for folks too.

jsantell:
I'd love to add webpack imports to make this nicer since #241 landed, but that'd require others importing via webpack to also need the same webpack configuration, although it'd make a nicer experience for those just using the built script.

jsantell:
With WakeLock being a dependency, and rotate instructions now [moved to its own file](https://github.com/googlevr/cardboard-vr-display/commit/73df574c9c61d73a10b81e308bc0d13e19bf44d9), this looks solved. Looking at ways to reducing this filesize as well while addressing #210

PR #238, opened 2017-05-25
bolt on window vars to global in node

kellyselden:
This will make it easier to run in node, and make the following code https://github.com/kellyselden/ember-aframe/blob/13a67e3d29d7af6f5d01f20ed1e5722096d38f6b/index.js#L16-L17 and https://github.com/aframevr/aframe/blob/cafbadc493701ebc5da33e531ea490e20a7eaa88/tests/node/test.js#L11-L12 obsolete.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "watchify": "latest",
     "uglify-js": "^3.0.11"
   },
-  "main": "build/webvr-polyfill.js",
+  "main": "src/node-entry",
   "keywords": [
     "vr",
     "webvr"
--- a/src/node-entry.js
+++ b/src/node-entry.js
@@ -1,0 +1,8 @@
+// if running in node and there is a window mock available,
+// globalize its members where needed
+if (global && global.window) {
+  global.document = global.window.document;
+  global.navigator = global.window.navigator;
+}
+
+require('./main');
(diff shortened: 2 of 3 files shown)


jsantell:
This looks good as well, could you fix the conflicts from #236?

kellyselden:
Rebased.

jsantell:
Thanks! I think in the future, we should move back to exposing just the distributed built file even for node for reasons like supporting transformations (currently only JSON, but potentially others like #120), without requiring the consumer of this module to add them as well, but also have better support for Node inclusions from the get-go. In the mean time, happy to support this fix. Thanks!

kellyselden:
Sounds good. I didn't think you wanted unused node-only code to make it down to the browser, that's why extracted it to a new entry file.

AlbertoElias:
With this change, compiling for frontend is now a bit harder as tooling use the main key in package.json, but then running it in the browser breaks as that is Node syntax

jsantell:
@AlbertoElias  I think changing `if (typeof global !== 'undefined' && global.window)` will fix this issue -- sorry about that, working on tests to prevent this sort of thing from happening! I'll land a fix shortly.

AlbertoElias:
Yes, that would fix it :)

jsantell:
@AlbertoElias this should be fixed in 0.9.34

AlbertoElias:
Thanks! 👌

PR #239, opened 2017-05-25
bundle jsdom for window mocking

kellyselden:
This is the final step to make running this polyfill in node a seemless experience. Previously, you would have to supply your own `window` implementation (jsdom) if you wanted to consume this. After this change, you get jsdom's window mocking for free, and all you have to do to use this is `require('webvr-polyfill')`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,13 +10,13 @@
   "description": "Use WebVR today, on mobile or desktop, without requiring a special browser build.",
   "dependencies": {
     "eventemitter3": "^2.0.2",
+    "jsdom": "^9.12.0",
     "object-assign": "^4.0.1"
   },
   "devDependencies": {
     "browserify": "latest",
     "chai": "^3.5.0",
     "derequire": "latest",
-    "jsdom": "^9.12.0",
     "mocha": "^3.2.0",
     "watchify": "latest",
     "uglify-js": "^3.0.11"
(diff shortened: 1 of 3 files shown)


jsantell:
This would increase the size of browserify/webpack includes (which would use src/node-entry) from 244k to 5.1M (!), in addition to possibly conflicting (duplicate) versions of jsdom for a node app using jsdom. I think the solution to this is **related** to #47, and think it should take a window object (defaulting to `window`), where a jsdom window could be provided.

Providing alternative DOMs is out of the scope of this project, especially with the code size increase in browser scenarios, but **supporting** alternative DOMs, via passing in a global, should be supported. Would that cover your use case @kellyselden ?

kellyselden:
I'm a little confused. The browserify bundler wouldn't use the `src/node-entry` file would it? The build script uses `browserify src/main.js` as the entry point. So I don't see how the bundle size would explode like that, unless browserify automatically bundles everything in npm dependencies?

jsantell:
Correct, but if aframe or someone else using this for example calls in `require('webvr-polyfill')`, THEIR browserify/webpack/loader will call whatever is in `main` in the package.json and build the polyfill, rather than use the `build/webvr-polyfill.js` due to #238

kellyselden:
Ah you're right. I'm a little rusty on bundlers.

jsantell:
@kellyselden would exposing the polyfill to accept a window/jsdom object work for your use case? Something like `new WebVRPolyfill(window, config)`?

kellyselden:
Yes it would.

PR #244, opened 2017-05-30
Fire 'vrdisplayconnect' event when populating VRDisplays. Fixes #40

### LINKED ISSUE #40: Emit `vrdisplayconnected` and `vrdisplaydisconnected` events
The [events](https://mozvr.github.io/webvr-spec/#interface-window) fired on `window`:
- `vrdisplayconnected` (not implemented in latest Chromium builds)
- `vrdisplaydisconnected` (not implemented)
- `vrdisplaypresentchange` (implemented)

jsantell:
Note, `vrdisplaydisconnect` doesn't apply here as there's no way to remove a shim VRDisplay anyway.

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -342,6 +342,14 @@ VRDisplay.prototype.fireVRDisplayPresentChange_ = function() {
   window.dispatchEvent(event);
 };
 
+VRDisplay.prototype.fireVRDisplayConnect_ = function() {
+  // Important: unfortunately we cannot have full spec compliance here.
+  // CustomEvent custom fields all go under e.detail (so the VRDisplay ends up
+  // being e.detail.display, instead of e.display as per WebVR spec).
+  var event = new CustomEvent('vrdisplayconnect', {detail: {display: this}});
+  window.dispatchEvent(event);
+};
+
 VRDisplay.prototype.addFullscreenListeners_ = function(element, changeHandler, errorHandler) {
   this.removeFullscreenListeners_();
 
--- a/src/webvr-polyfill.js
+++ b/src/webvr-polyfill.js
@@ -54,6 +54,11 @@ WebVRPolyfill.prototype.isDeprecatedWebVRAvailable = function() {
   return ('getVRDevices' in navigator) || ('mozGetVRDevices' in navigator);
 };
 
+WebVRPolyfill.prototype.connectDisplay = function(vrDisplay) {
+  vrDisplay.fireVRDisplayConnect_();
+  this.displays.push(vrDisplay);
+};
+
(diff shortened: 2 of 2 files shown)


toji:
LGTM

src/base.js
```diff
@@ -342,6 +342,14 @@ VRDisplay.prototype.fireVRDisplayPresentChange_ = function() {
   window.dispatchEvent(event);
 };
 
+VRDisplay.prototype.fireVRDisplayConnect_ = function() {
+  // Important: unfortunately we cannot have full spec compliance here.
+  // CustomEvent custom fields all go under e.detail (so the VRDisplay ends up
+  // being e.detail.display, instead of e.display as per WebVR spec).
+  var event = new CustomEvent('vrdisplayconnect', {detail: {display: this}});
```
cvan: also another note: (FYI: this probably isn't an issue because the Polyfill realistically probably doesn't need to support IE 11.)

Edge is fine, but in IE 11 `CustomEvent` isn't exposed as a constructor (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill)). so you have to use `document.createElement` and `initCustomEvent`, like so:

```js
var event = document.createEvent('CustomEvent');
event.initCustomEvent(…);
```

same as [here](https://github.com/googlevr/webvr-polyfill/pull/244/files#diff-d7e697eff3913f1acaacac8002c3b05eR341) and probably elsewhere too.

src/webvr-polyfill.js
```diff
@@ -54,6 +54,11 @@ WebVRPolyfill.prototype.isDeprecatedWebVRAvailable = function() {
   return ('getVRDevices' in navigator) || ('mozGetVRDevices' in navigator);
 };
 
+WebVRPolyfill.prototype.connectDisplay = function(vrDisplay) {
+  vrDisplay.fireVRDisplayConnect_();
```
cvan: FYI: last time I checked, Firefox, Chrome, and Chromium don't fire `vrdisplayconnect` when the page loads. only when a device is manually plugged in/reconected. that still make sense in this case?

src/base.js
jsantell: FWIW, I was going off of the spec for `vrdisplayconnect` -- and good call on IE, I'll add a follow up issue for that, agreed low priority due to IE11, filed #249

src/webvr-polyfill.js
jsantell: That's my understanding as well, but this handled from the perspective of these shim displays don't exist/aren't connected until they're created. Also probably low usage using these events in polyfill for that very reason (these can't be "connected"/"disconnected" in the middle of a session)

PR #248, opened 2017-05-31
Set a timeout on getVRDisplays() to handle scenario where the promise never resolves. Fixes #199

### LINKED ISSUE #199: Using native WebVR API without Google VR Services installed
When I add the Origin trial token on a project, with public Chrome 56 it works OK on Daydream ready phones (like Pixel) it returns "Google, Inc. Daydream View” on VRDisplay and I can enter on Daydream mode. But when I tried the same project with a Nexus5 or a Samsung S6 or Note7 (non-daydream ready phones) VRDisplay returns "Cardboard VRDisplay (webvr-polyfill)” which seems to be correct, but if I have the meta of the WebVR token on the header, the camera doesn’t move with the accelerometer. But if I comment the meta, it works perfect. (On Chrome 55 worked fine)

I downloaded last version of the repo and I have upload to my server (which have https and Origin trial approved for this domain) two htmls, one similar to the basic example (without the meta) and another with the origin trial meta added.
[https://unboring.net/tmp/webvr-polyfill/](https://unboring.net/tmp/webvr-polyfill/)

With a Google Pixel (Android 7.1.1) + Chrome 56 >
- If you open [with Origin trial added](https://unboring.net/tmp/webvr-polyfill/basic/with.html) version.  **(OK)**
vrDisplay.displayName = "Google, Inc. Daydream View" > VR button go to Daydream mode.

- If you open [without Origin trial](https://unboring.net/tmp/webvr-polyfill/basic/) version. **(OK)**
vrDisplay.displayName = "Cardboard VRDisplay (webvr-polyfill)" > VR button go to Cardboard mode.

With a Nexus5/Samsung S6 (Android 6.0.1) + Chrome 56 >
- If you open [with Origin trial added](https://unboring.net/tmp/webvr-polyfill/basic/with.html) version. **BUG. Doesn't move the camera with the accelerometer.**
vrDisplay.displayName = "Cardboard VRDisplay (webvr-polyfill)" > VR button go to Cardboard mode.

- If you open [without Origin trial](https://unboring.net/tmp/webvr-polyfill/basic/) version. **(OK)**
vrDisplay.displayName = "Cardboard VRDisplay (webvr-polyfill)"> VR button go to Cardboard mode.

jsantell:
Still need to do some testing here to make sure this is the appropriate fix

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -68,7 +68,15 @@ window.WebVRConfig = Util.extend({
   // When set to true, this will cause a polyfilled VRDisplay to always be
   // appended to the list returned by navigator.getVRDisplays(), even if that
   // list includes a native VRDisplay.
-  ALWAYS_APPEND_POLYFILL_DISPLAY: false
+  ALWAYS_APPEND_POLYFILL_DISPLAY: false,
+
+  // There are scenarios where the native WebVR API exists, and instead of returning
+  // 0 VR displays when none are detected, `navigator.getVRDisplays()`'s promise never
+  // resolves. This results in the polyfill hanging and not being able to provide fallback
+  // displays, so set a timeout in milliseconds to stop waiting for a response
+  // and just use polyfilled displays.
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=727969
+  GET_VR_DISPLAYS_TIMEOUT: 1000,
 }, window.WebVRConfig);
 
 if (!window.WebVRConfig.DEFER_INITIALIZATION) {
--- a/src/webvr-polyfill.js
+++ b/src/webvr-polyfill.js
@@ -153,23 +153,29 @@ WebVRPolyfill.prototype.getVRDisplays = function() {
   this.populateDevices();
   var polyfillDisplays = this.displays;
 
-  if (this.nativeWebVRAvailable) {
-    return this.nativeGetVRDisplaysFunc.call(navigator).then(function(nativeDisplays) {
-      if (window.WebVRConfig.ALWAYS_APPEND_POLYFILL_DISPLAY) {
-        return nativeDisplays.concat(polyfillDisplays);
-      } else {
-        return nativeDisplays.length > 0 ? nativeDisplays : polyfillDisplays;
-      }
-    });
-  } else {
-    return new Promise(function(resolve, reject) {
-      try {
-        resolve(polyfillDisplays);
-      } catch (e) {
-        reject(e);
-      }
-    });
+  if (!this.nativeWebVRAvailable) {
+    return Promise.resolve(polyfillDisplays);
   }
+
+  // Set up a race condition if this browser has a bug where
+  // `navigator.getVRDisplays()` never resolves.
+  var vrDisplaysNative = this.nativeGetVRDisplaysFunc.call(navigator);
+  var timeout = new Promise(function(resolve) {
+    setTimeout(function() {
+      resolve([])
+    }, window.WebVRConfig.GET_VR_DISPLAYS_TIMEOUT);
+  }).then(function (displays) {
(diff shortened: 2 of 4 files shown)


jsantell:
More information #199 and https://bugs.chromium.org/p/chromium/issues/detail?id=727969¬

jsantell:
Testing with a local fork of this [webvr.info samples](https://webvr.info/samples/03-vr-presentation.html?polyfill=1) I can see it fail currently with the webvr chrome://flags enabled (same with using origin trial, but this was local), and when building with this patch, it works. There's a 1000ms (configurable) long gap from loading and when it "snaps" into receiving the orientation controls due to the delay waiting for a response from `getVRDisplays()`, and we could make it lower, but rather be safe in this case.

I do still get the "Install Google VR Services" pop up in this case, but this handles the issue where it's not installed, old, or unavailable via https://github.com/googlevr/webvr-polyfill/pull/248#issuecomment-305329755, and the polyfill never kicks in, halting all screen movement.

jsantell:
This is a weird, but important one; pinging @toji and @cvan for review

toji:
LGTM. Thanks for handling this edge case.

src/webvr-polyfill.js
```diff
@@ -153,23 +153,29 @@ WebVRPolyfill.prototype.getVRDisplays = function() {
...
+
+  // Set up a race condition if this browser has a bug where
+  // `navigator.getVRDisplays()` never resolves.
+  var vrDisplaysNative = this.nativeGetVRDisplaysFunc.call(navigator);
+  var timeout = new Promise(function(resolve) {
+    setTimeout(function() {
+      resolve([])
+    }, window.WebVRConfig.GET_VR_DISPLAYS_TIMEOUT);
+  }).then(function (displays) {
+    console.warn('Native WebVR implementation detected, but `getVRDisplays()` failed to resolve. Falling back to polyfill.');
```
tforcram: Couldn't this warn be part of the setTimeout handler?  The extra then handler here (which takes and just returns displays again) feels unneeded.

src/main.js
```diff
@@ -68,7 +68,15 @@ window.WebVRConfig = Util.extend({
   // When set to true, this will cause a polyfilled VRDisplay to always be
   // appended to the list returned by navigator.getVRDisplays(), even if that
   // list includes a native VRDisplay.
-  ALWAYS_APPEND_POLYFILL_DISPLAY: false
+  ALWAYS_APPEND_POLYFILL_DISPLAY: false,
+
+  // There are scenarios where the native WebVR API exists, and instead of returning
```
tforcram: This shouldn't be true in the long term, so maybe refer to this as handling the possibility of bugs in the implementation instead of 'scenarios' which seem to imply the promise not returning can be expected in some cases.

src/webvr-polyfill.js
jsantell: @tforcram Both solutions for the message aren't working as expected, since the timeout will always resolve, but it's up to the `Promise.race` to determine which finished first, checking it out

src/main.js
jsantell: @tforcram Good point, I'll clarify this with specific ranges and not something that will occur Chrome M60+

src/webvr-polyfill.js
jsantell: To clarify, the warning message will always display which is an issue -- the fallback functionality works as expected

tforcram: Ah, yes that would happen with race because it just lets both promises run (and doesn't try to cancel the ones that lost).

You would need a shared variable to mark as complete which you could check in the resolver of the losing function to have it end early.

tforcram: Or probably cleaner to keep save the timer id and then call clearTimeout on it from the native displays promise if it finishes first.

PR #259, opened 2017-06-08
Expose InstallWebVRSpecShim

huningxin:
To fix https://github.com/toji/webvr.info/issues/63

### DIFF
--- a/src/webvr-polyfill.js
+++ b/src/webvr-polyfill.js
@@ -272,6 +272,7 @@ function InstallWebVRSpecShim() {
   }
 };
 
+WebVRPolyfill.InstallWebVRSpecShim = InstallWebVRSpecShim;
 WebVRPolyfill.version = version;
 
 module.exports.WebVRPolyfill = WebVRPolyfill;
--- a/README.md
+++ b/README.md
@@ -115,6 +115,11 @@ Performance is critical for VR. If you find your application is too sluggish,
 consider tweaking some of the above parameters. In particular, keeping
 `BUFFER_SCALE` at 0.5 (the default) will likely help a lot.
 
+## WebVR 1.1 Shim
+
+The polyfill exposes a helper method `WebVRPolyfill.InstallWebVRSpecShim` which
+installs a shim that updates a WebVR 1.0 spec implementation to WebVR 1.1.
+
 ## Development
 
 If you'd like to contribute to the `webvr-poyfill` library, check out


huningxin:
@jsantell @toji , PTAL. Thanks!

src/main.js
```diff
@@ -89,3 +90,4 @@ if (!window.WebVRConfig.DEFER_INITIALIZATION) {
 }
 
 window.WebVRPolyfill = WebVRPolyfill;
+window.InstallWebVRSpecShim = InstallWebVRSpecShim;
```
cvan: can this instead be a method exposed as a method on `WebVRPolyfill`?

src/webvr-polyfill.js
```diff
@@ -275,3 +275,4 @@ function InstallWebVRSpecShim() {
 WebVRPolyfill.version = version;
 
 module.exports.WebVRPolyfill = WebVRPolyfill;
+module.exports.InstallWebVRSpecShim = InstallWebVRSpecShim;
```
cvan: same here

cvan:
thanks for opening this PR. this is a good idea to expose the method. if you could add document this in the [`README.md`](https://github.com/googlevr/webvr-polyfill/blob/master/README.md#configuration) too, that'd be great.

jsantell:
:+1: on attaching it to `WebVRPolyfill.installWebVRSpecShim` -- I don't think it makes sense as a prototype method, as instantiating the polyfill calls it anyway (and strange to have to manually call `WebVRPolyfill.prototype.installWebVRSpecShim()`.

src/main.js
huningxin: sounds good. Will do that.

src/webvr-polyfill.js
huningxin: acked.

huningxin:
Updated the PR according to your comments. @cvan @jsantell , please take another look.

huningxin:
@cvan @jsantell , any comments?

jsantell:
@huningxin  Looks good to me! Could you squash this down to one commit and I'll merge it in?

toji:
LGTM as well. GitHub can do the squash and merge for us, any reason not to go that route?

huningxin:
Thanks for review! And the commits have been squashed. Please help merge. Thanks!

jsantell:
Thanks @huningxin! Merging in

PR #264, opened 2017-06-20
fix webvr polyfill for chromium/rift (fixes #263)

### LINKED ISSUE #263: webvr-polyfill 0.9.34 breaks on Chromium Desktop with Rift
1. Use Chromium June build on Windows.
2. Reproduced on A-Frame master using 0.9.34.
3. Try to enter VR, only enters full screen, and Rift displays "Please wait..."

Reverting aframevr/aframe#ebf12fff74a3323598fb4c34b9144ed86725ae77 to the previous branch of the polyfill removes this issue. So we have isolated the issue to the polyfill.

ngokevin:
Reverts https://github.com/googlevr/webvr-polyfill/commit/1ac010e67ffa50876ff187a647460784bf158d1d

Not sure the details, but this fixes Chromium (June) + Rift on Windows Desktop for me. Open for discussion on better ways to fix since I'm not sure what the offending patch was intending.


ngokevin:
I signed it!

src/webvr-polyfill.js
```diff
@@ -107,7 +107,7 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
   navigator.getVRDisplays = this.getVRDisplays.bind(this);
 
   // Polyfill native VRDisplay.getFrameData
-  if (this.nativeWebVRAvailable && window.VRFrameData) {
+  if (this.nativeWebVRAvailable && this.isCardboardCompatible() && window.VRFrameData) {
```
dmarcos: Does this have any side effects when the condition is legitimate? Should the fix go in the implementation of `isCardboardCompatible` so it does not give false positives / negatives?

dmarcos: I understand better now. Should then the computation of `this.nativeWebVRAvailable` consider the value  of `this.isCardboardCompatible()`?

dmarcos: Why is the API polyfilled if it exists?

cvan: @dmarcos: @ngokevin: see the discussion in https://github.com/googlevr/webvr-polyfill/issues/177#issuecomment-282671833

@brianchirls: do you have thoughts now on a better way to feature-detect for this? or perhaps expose this ["polyfilled" `VRFrameData`](https://github.com/googlevr/webvr-polyfill/pull/264/files#diff-7bfb91b61bb80cff41aba297525c4dc7R144) under a different name?

brianchirls: I'm not sure why this broke things. I'll take a look...

No, I don't think `nativeWebVRAvailable` should consider `isCardboardCompatible()`.

The reason `VRFrameData` is polyfilled even if it exists is explained here:
https://github.com/googlevr/webvr-polyfill/pull/197

ngokevin:
I've tested this PR using A-Frame and this build of the polyfill on:

- [x] iOS Safari, iPhone 6
- [x] HTC Vive, Firefox Nightly
- [x] Oculus Rift, Brandon's Chromium June

brianchirls:
@ngokevin Please hold off on this PR. I think I have an idea of what's really causing this problem, but I don't have time to write it up right now. Will do it tonight or in the morning.

In the meantime, can you tell me if the bug happens on the unpatched polyfill with the Vive as well as the Rift?

It seems to me reverting that commit would break mouse and keyboard controls, no?

jsantell:
I logged a bunch more info in #266 -- serves me right for not checking my bugmail today :cry:

ngokevin:
Thanks! I will try your patch, thanks for finding the root issue.

brianchirls:
@ngokevin Cool. Looking forward to your feedback and testing results.

PR #267, opened 2017-06-22
Skip polyfilled `VRFrameData` if object passed to `getFrameData` is native

### LINKED ISSUE #263: webvr-polyfill 0.9.34 breaks on Chromium Desktop with Rift

brianchirls:
To operate properly, the polyfill needs to load and run as early as possible. Even if WebVR API is available, `VRFrameData` is polyfilled in case the browser provides no displays and we need to create a polyfilled Cardboard display.

This patch fixes the case where a native `VRFrameData` instance is created before the polyfill can be applied. `VRDisplay.prototype.getFrameData` now checks its argument so it doesn't try to set the read-only `pose` property on a native object.

I've also taken the liberty of adding a reference to the original un-polyfilled `VRFrameData` as `WebVRPolyfill.prototype.NativeVRFrameData` in case someone wants to reference it.

Fixes #263

### DIFF
--- a/src/webvr-polyfill.js
+++ b/src/webvr-polyfill.js
@@ -108,11 +108,17 @@ WebVRPolyfill.prototype.enablePolyfill = function() {
 
   // Polyfill native VRDisplay.getFrameData
   if (this.nativeWebVRAvailable && window.VRFrameData) {
+    var NativeVRFrameData = window.VRFrameData;
     var nativeFrameData = new window.VRFrameData();
     var nativeGetFrameData = window.VRDisplay.prototype.getFrameData;
     window.VRFrameData = VRFrameData;
 
     window.VRDisplay.prototype.getFrameData = function(frameData) {
+      if (frameData instanceof NativeVRFrameData) {
+        nativeGetFrameData.call(this, frameData);
+        return;
+      }
+
       /*
       Copy frame data from the native object into the polyfilled object.
       */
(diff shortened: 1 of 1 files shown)


jsantell:
Looks like a clean patch, thanks for handling this @brianchirls! I won't be able to test this out until later and get back to you

machenmusik:
@dmarcos @ngokevin @brianchirls @jsantell 
quick test results with latest 20170607 Chromium and Rift -
(1) when I put on the Rift with it outside tracking area, it crashed the tab - not sure if related
(2) when I reloaded the tab and put the Rift on while inside the tracking area, worked great!

dmarcos:
I confirm that this seems to fix the issue for me on Chromium 06/06/2017 and Oculus on Windows 10 with a GTX 980. I have a local build of aframe.io with this PR as the polyfill and I can visit the examples and enter VR

dmarcos:
Confirmed that this PR works with Vive / Windows 10 / Nightly

RSpace:
Works for me with Rift / Windows 10 / Chromium. Thanks!

jsantell:
Thanks all! This has been published in 0.9.35

PR #270, opened 2017-06-30
Fix fusion-pose-sensor.js for Firefox on Android

bluemarvin:
Firefox for android returns the timestamp correctly now so there is no need to divide by 1000 any more.


bluemarvin:
I signed it

jsantell:
Thanks @bluemarvin! Missed the notification for this, sorry for the delay!

Is this a change in Firefox Nightly/56? Should this also check something else to work in Firefox < 56, as is currently?

bluemarvin:
I'm not sure which version of Fennec the fix landed in, but it is in current release so I don't know that it is worth checking versions.

jsantell:
Confirmed with Firefox for Android release 54 that this fixes it -- thanks @bluemarvin for the PR!

jsantell:
Landed in v0.9.36, thanks!

PR #276, opened 2017-08-17
only run polyfill if nativeWebVR is not available

gkatsev:
If you run the polyfill when native VR is available, you won't be able to put the player into VR mode properly. This should fix https://github.com/borismus/webvr-boilerplate/issues/170 and https://github.com/videojs/videojs-vr/issues/14


cvan:
pretty good catch. @jsantell: look good to you?

jsantell:
This looks good, thanks for this!

gkatsev:
You're welcome!

PR #277, opened 2017-08-17
prevent redefinition of navigator.vrEnabled

mikestaub:
This tripped me up today. I don't think we want to redefine the property if it is `false`. If we do, then we should instead set `writable: true` in defineProperty()'s third argument.


mikestaub:
I signed it!

jsantell:
Thanks for this, looks good!

cvan:
@mikestaub sorry, didn't first see your PR here before merging #281. in any event, it's fixed now. many thanks for the catch and for the PR 👍

Issue #280, opened 2017-08-29
Polyfill fails on current Chrome and Safari on macOS High Sierra


joshmarinacci:
I get an error about not being able to modify the navigator.vrEnabled property.  The root cause is that the code assumes if !navigator.vrEnabled then it should add the property. This doesn't consider if the property *is* defined and was explicitly set to false.

cvan:
is #281 the only fix needed? can you test and report back? thanks for filing this issue and your patience 👍

joshmarinacci:
Yep.

jsantell:
Thanks for filing! Looks like #281 fixes this. Will cut a release later today with these changes.

PR #281, opened 2017-08-29
fix for github issue #280

joshmarinacci:
Don't try to modify navigator.vrEnabled if it does exist but is set to false. Check for undefined instead.  fix for github issue #280


joshmarinacci:
I signed it!

cvan:
good catch, thanks

Issue #282, opened 2017-09-06
WebVR-polyfill with Unity WebGL publish


themolecule:
Couldn't get it working with any form of Unity WebGLTemplate in iOS Safari until I changed lines 3529, 3530, and 3533:

~~~~
CardboardVRDisplay.prototype.onResize_ = function(e) {                                  
  if (this.layer_) {                                                                    
    var gl = this.layer_.source.getContext('webgl');                                    
    // Size the CSS canvas.                                                             
    // Added padding on right and bottom because iPhone 5 will not                      
    // hide the URL bar unless content is bigger than the screen.                       
    // This will not be visible as long as the container element (e.g. body)            
    // is set to 'overflow: hidden'.                                                    
    // Additionally, 'box-sizing: content-box' ensures renderWidth = width + padding.   
    // This is required when 'box-sizing: border-box' is used elsewhere in the page.    
    var cssProperties = [                                                               
      'position: absolute',                                                             
      'top: 0',                                                                         
      'left: 0',                                                                        
      'width: 100vw',             //3529 use vw, not pixels (which don't account for devicePixelRatio
      'height: 100vh',            //3530 use vw, not pixels (which don't account for devicePixelRatio
      'border: 0',                                                                      
      'margin: 0',                                                                      
      'padding: 0px',             //3533 remove padding
      'box-sizing: content-box',                                                        
    ];                                                                                  
    gl.canvas.setAttribute('style', cssProperties.join('; ') + ';');                    
                                                                                        
    Util.safariCssSizeWorkaround(gl.canvas);                                            
  }                                                                                     
};                                                    
~~~~

Because Unity 2017.x wants to receive a div, not a canvas, to grab onto, we don't have clean control over where the canvas sits in the DOM... the padding was causing miscalculations for the canvas size, which was causing it to grow each frame until the browser crashes.

PR #284, opened 2017-09-12
Use viewport sizes instead of pixels, and remove padding for cardboard VR display

jsantell:
via #282 

Need to test on an iOS device


dmotz:
This PR is great and fixes all the issues I'm having with resizing on iOS. Any chance of a merge/npm update soon? Thanks!

PR #285, opened 2017-10-10
use webpack-dev-server 2.7.1 to fix `npm start`

machenmusik:
https://github.com/webpack/webpack-dev-server/issues/1101#issue-258321001

PR #286, opened 2017-10-10
Make polyfill work with ODG R-7 Web Browser

machenmusik:
The changes here make the accelerometer / gyroscope based sensor fusion work properly on the latest Web Browser for the ODG R-7, which is a landscape-primary device.

### DIFF
--- a/src/sensor-fusion/fusion-pose-sensor.js
+++ b/src/sensor-fusion/fusion-pose-sensor.js
@@ -139,7 +139,11 @@ FusionPoseSensor.prototype.updateDeviceMotion_ = function(deviceMotion) {
     this.previousTimestampS = timestampS;
     return;
   }
+
   this.accelerometer.set(-accGravity.x, -accGravity.y, -accGravity.z);
+  if (Util.isR7()) {
+    this.gyroscope.set(-rotRate.beta, rotRate.alpha, rotRate.gamma);
+  } else
   this.gyroscope.set(rotRate.alpha, rotRate.beta, rotRate.gamma);
 
   // With iOS and Firefox Android, rotationRate is reported in degrees,
--- a/src/sensor-fusion/fusion-pose-sensor.js
+++ b/src/sensor-fusion/fusion-pose-sensor.js
@@ -143,8 +143,9 @@ FusionPoseSensor.prototype.updateDeviceMotion_ = function(deviceMotion) {
   this.accelerometer.set(-accGravity.x, -accGravity.y, -accGravity.z);
   if (Util.isR7()) {
     this.gyroscope.set(-rotRate.beta, rotRate.alpha, rotRate.gamma);
-  } else
-  this.gyroscope.set(rotRate.alpha, rotRate.beta, rotRate.gamma);
+  } else {
+    this.gyroscope.set(rotRate.alpha, rotRate.beta, rotRate.gamma);
+  }
 
   // With iOS and Firefox Android, rotationRate is reported in degrees,
   // so we first convert to radians.
--- a/src/util.js
+++ b/src/util.js
@@ -80,8 +80,16 @@ Util.isFirefoxAndroid = (function() {
   };
 })();
 
+Util.isR7 = (function() {
+  var isR7 = navigator.userAgent.indexOf('R7 Build') !== -1;
+  return function() {
+    return isR7;
+  };
+})();
+
 Util.isLandscapeMode = function() {
-  return (window.orientation == 90 || window.orientation == -90);
+  var rtn = (window.orientation == 90 || window.orientation == -90);
+  return Util.isR7() ? !rtn : rtn;
 };
 
 // Helper method to validate the time steps of sensor timestamps.


src/util.js
```diff
@@ -80,8 +80,16 @@ Util.isFirefoxAndroid = (function() {
   };
 })();
 
+Util.isR7 = (function() {
```
jsantell: This could be more future-proof by generalizing it as `isLandscapePrimary()` for similar devices in the future that are not R7's. Also some comments on what landscape primary means (orientation off by 90?). Is there any other way of detecting a landscape primary device in the navigator other than user agent parsing?

package.json
```diff
@@ -14,7 +14,7 @@
     "mocha": "^3.2.0",
     "semver": "^5.3.0",
     "webpack": "^2.6.1",
-    "webpack-dev-server": "^2.4.5"
+    "webpack-dev-server": "2.7.1"
```
jsantell: Let's keep this change in #285

jsantell:
Cool, not familiar with how the R7 works -- is the orientation always off by 90 compared to (portrait) mobile devices? Added some comments inline; if those are addressed, this looks good to land -- thanks!

src/util.js
machenmusik: On the one hand, I agree.  On the other hand, I'm not entirely sure whether the behavior observed with the R-7 is correct for a landscape-primary device and is the right basis to generalize.  Are there others you know of that can be used to confirm?

package.json
machenmusik: oops, I was testing with `npm start`, will remove.

machenmusik:
The basic example works with R-7, but then using it with A-Frame shows the whole display rotated 90 degrees (courtesy of THREE.VRControls), so maybe the way it is solved here is not the best path -- thoughts?

machenmusik:
On the R-7, the default view is landscape, and window.orientation reports 0 (since the device is not rotated).

src/util.js
jsantell: Unfortunately no, this configuration for a browser is new and surprising to me :smile:

machenmusik: I don't think that `window.screen.orientation.type` is generally implemented either, which seems like the "right" way to indicate and check.

jsantell: What do CSS media queries report for portrait/landscape?

machenmusik: not sure, is there a webpage I can load and tell you what it says?

machenmusik:
OK, this latest version doesn't have the problem when subsequently used with A-Frame, and shouldn't have the build files or webpack-dev-server change.

src/util.js
jsantell: Made this quickly: https://heavy-tempo.glitch.me/, just renders the background differently from media queries, but appears that these [don't necessarily correlate to device orientation..](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation).

```
<html>
  <head>
    <style>
      @media screen and (orientation: landscape) {
        body { background-color:red; } }
      @media screen and (orientation: portrait) {
        body { background-color: blue; } }
    </style>
  </head>
  <body>
  </body>
</html>
```

Also it seems that `orientation.type`, even if implemented, is just a string version of the rotation in degrees, so that would change on all devices on every rotation. Does the R7 at least implement orientation.type? Maybe a general getOrientation function would be useful, returning the orientation in degrees, and attempts the orientation type values (again, hoping that landscape-primary devices [which is different than the [landscape-primary value from orientation-type]( https://developer.mozilla.org/en-US/docs/Web/API/Screen/lockOrientation)] implement this)

src/sensor-fusion/fusion-pose-sensor.js
```diff
@@ -139,7 +139,11 @@ FusionPoseSensor.prototype.updateDeviceMotion_ = function(deviceMotion) {
     this.previousTimestampS = timestampS;
     return;
   }
+
   this.accelerometer.set(-accGravity.x, -accGravity.y, -accGravity.z);
+  if (Util.isR7()) {
+    this.gyroscope.set(-rotRate.beta, rotRate.alpha, rotRate.gamma);
+  } else
```
jsantell: small change: curly bracket for the `else` statement and same indentation as the `if`

```
} else {
  this.gyroscope.set(rotRate.alpha, rotRate.beta, rotRate.gamma);
}
```

machenmusik:
Rebase worked correctly this time.  OK to merge?

src/util.js
machenmusik: The latest version of the change doesn't mess with window.orientation (which I think is what messed up A-Frame down the line), but simply modifies the interpretation of the gyroscope.

Until we understand more about other landscape-primary devices, and how we can reliably detect them, I think it may be better to keep this scoped to R-7 and expand/generalize in subsequent PRs.

machenmusik:
done, thanks!

jsantell:
Looks good, thanks @machenmusik!

PR #295, opened 2017-11-25
Add the option to change the tracked container

Aykelith:
I made some changes that I need because I did not wanted the `window` to be tracked, but another container.

### DIFF
--- a/src/mouse-keyboard-vr-display.js
+++ b/src/mouse-keyboard-vr-display.js
@@ -34,11 +34,16 @@ function MouseKeyboardVRDisplay() {
 
   this.capabilities.hasOrientation = true;
 
+  this.onKeyDown_ = this.onKeyDown_.bind(this);
+  this.onMouseDown_ = this.onMouseDown_.bind(this);
+  this.onMouseMove_ = this.onMouseMove_.bind(this);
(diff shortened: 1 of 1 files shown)


Aykelith:
I signed it!

jsantell:
Thanks for the PR @Aykelith! Let me think for a bit on this, as I'm hesitant to add too many non-VRDisplay functions to one of the implementing classes and something that could be done outside of the polyfill. I wonder if there's another solution that makes this customizable by the consumer, like bind the event handlers to the MouseKeyboardVRDisplay like you have (so that they may be unbound later):

```js
this.onKeyDown_ = this.onKeyDown_.bind(this);
this.onMouseDown_ = this.onMouseDown_.bind(this);
this.onMouseMove_ = this.onMouseMove_.bind(this);
this.onMouseUp_ = this.onMouseUp_.bind(this);
```

And then in consumer code:
```js
window.removeEventListener('mouseup', vrDisplay.onMouseUp_);
// Remove other handlers on window
container.addEventListener('mouseup', vrDisplay.onMouseUp_);
// Add other handlers on container
```
Would something like that also solve the issue you're facing?

Aykelith:
Yes, I changed the code and kept just the bindings.

jsantell:
Thank you, @Aykelith!

PR #296, opened 2017-12-07
Re-add Util.getQuaternionAngle to fix debug mode

MattiasBuelens:
When debug mode is activated on mobile devices, the polyfill throws a lot of errors:
```
Uncaught TypeError: Util.getQuaternionAngle is not a function
    at ComplementaryFilter.run_ (webvr-polyfill.js:4237)
    at ComplementaryFilter.addGyroMeasurement (webvr-polyfill.js:4190)
    at FusionPoseSensor.updateDeviceMotion_ (webvr-polyfill.js:4039)
    at FusionPoseSensor.onDeviceMotion_ (webvr-polyfill.js:4009)
```
Steps to reproduce:
1. Open http://googlevr.github.io/webvr-polyfill/examples/basic/?debug=true in Chrome on Android
2. Attach a remote Chrome debugger
3. Click the VR button

It appears that `Util.getQuaternionAngle` was (accidentally) removed some time [after v0.2.6](https://github.com/googlevr/webvr-polyfill/blob/v0.2.6/fusion/util.js), but it was still used in debug mode. This PR re-adds this missing function.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -478,4 +478,15 @@ Util.getDomainFromUrl = function(url) {
   return domain;
 }
 
+Util.getQuaternionAngle = function(quat) {
+  // angle = 2 * acos(qw)
+  // If w is greater than 1 (THREE.js, how can this be?), arccos is not defined.
+  if (quat.w > 1) {
+    console.warn('getQuaternionAngle: w > 1');
+    return 0;
+  }
+  var angle = 2 * Math.acos(quat.w);
+  return angle;
+};
+
 module.exports = Util;


MattiasBuelens:
@googlebot I signed it

MattiasBuelens:
Perhaps off-topic, but what's the value of this debug mode? We accidentally ran into this because a customer was using `?debug` for debugging their own website, not the WebVR polyfill. If it's been broken for so long and nobody's actually using it, maybe it's better to remove it altogether?

jsantell:
Thanks for reporting this! It seems `debug` is somewhat vestigial, not one I ever used or dove into since I've started maintaining several months ago.

Long term, I think I'd rather remove this, or make it a value in the configuration, rather than a query param. In the mean time, let's merge this util back in. Thanks for digging into this!

